PR #5, opened 2018-05-22
Fix spelling of SUSE

thkukuk:
The lower 'u' in SUSE was gone 14 years ago.


besser82:
Thank you for your contribution!

PR #6, opened 2018-06-24
ALT changes potentially useful for other vendors migrating from glibc's libcrypt

ldv-alt:
This series is a bare minimum of changes necessary to migrate from ALT glibc's libcrypt.  Other vendors might find it useful, e.g. `--enable-weak-hashes=glibc --enable-obsolete-api=glibc` seems to be a good choice for Fedora.

PR #8, opened 2018-06-30
ALT changes potentially useful for other vendors migrating from glibc's libcrypt, part 1

ldv-alt:
This series is the first part of the changes that was necessary to migrate from ALT glibc's libcrypt.
This part contains those changes that were pre-approved in #6, and two commits with obvious fixes for bugs found later.


zackw:
Merged, thank you.

PR #9, opened 2018-06-30
ALT changes potentially useful for other vendors migrating from glibc's libcrypt, part 2

ldv-alt:
This series is the second part of the changes that was necessary to migrate from ALT glibc's libcrypt.
This part contains those changes that were not pre-approved in #6 and need to be discussed further.


ldv-alt:
What do you think about installing libcrypt.pc along with libxcrypt.pc, e.g. as a symlink?  This would achieve both goals.

besser82:
That would be an option.  Feel free to make the needed changes.  I'll review and merge then.  Please adapt `.travis.yml` to the changes to configure.

besser82:
Yay, LGTM now!

> I'd like to add new --enable-obsolete-api options to the test matrix,
but test-symbols-compat doesn't allow me to do that.

I think @zackw is able to help with that, as he wrote that particular test.

besser82:
I just checked the reason for `test-symbols-compat` to fail.  It is because Fedora already has libxcrypt and it defines those `OW_CRYPT_1.0` symbols.  Please and the different configure options to the test matrix and I'll merge, since I know the reason why some of them are failing.

zackw:
Off the top of my head, maybe test-symbols-compat should be skipped for any
non-default setting of --enable-obsolete-api. The point of the test is to
ensure binary backward compatibility with the system libcrypt, and if you
set that configure option you presumably don’t care about that.

besser82:
@zackw Well, that depends.  As I said:TravisCI compares against recent Fedora, which is build with `OWCRYPT` symbols, that's why it fails there until I've updated there and have built a version with `--enable-obsolete-api=glibc`.

ldv-alt:
On Tue, Jul 10, 2018 at 09:24:59AM -0700, Björn Esser wrote:
> @zackw Well, that depends.  As I said:TravisCI compares against recent Fedora, which is build with `OWCRYPT` symbols, that's why it fails there until I've updated there and have built a version with `--enable-obsolete-api=glibc`.

Something has to be done with the test, otherwise you won't be able
to build libxcrypt with --enable-obsolete-api=glibc to Fedora.

ldv-alt:
On Tue, Jul 10, 2018 at 06:08:42AM -0700, Zack Weinberg wrote:
> Off the top of my head, maybe test-symbols-compat should be skipped for any
> non-default setting of --enable-obsolete-api. The point of the test is to
> ensure binary backward compatibility with the system libcrypt, and if you
> set that configure option you presumably don’t care about that.

The use of e.g. --enable-obsolete-api=glibc (or --enable-obsolete-api=alt
in our case) doesn't mean that compatibility with the system libcrypt is
not important.  Compatibility shouldn't be broken accidentally.
test-symbols-compat should not be skipped unless an explicit removal
of some interfaces provided by system libcrypt is requested.

besser82:
LGTM!  Will rebase-merge.  Thank you for the enhancement!

PR #11, opened 2018-07-05
Make salt validation pickier.

zackw:
This started out as a patch to fold together test-crypt-badsalt.c and
test-crypt-nonnull.c (which were almost the same program) and extend
their testing from DES to all of the supported hashes.  That revealed
that many of the supported hash functions do not validate the contents
of their salt strings very carefully.

This patch has a low but nonzero backward compatibility risk, because
now we reject certain calls to crypt*() that we would previously have
accepted.  In particular, setting strings of the form

    $5$xxxxxxx*xxxxxxxx$

where x stands for any "itoa64" character and * for any non-"itoa64"
character, would formerly be accepted but are now rejected.  Some of
the hash algorithms that were lenient about the contents of the salt
would echo back the salt verbatim, and others would convert it to the
base64 alphabet somehow.  I think it's unlikely that this occurs in
real password files but I don't have a lot of data to base that on.

@besser82 Do you think this is a safe change to make?


besser82:
LGTM so far, but can you extend the `badsalt` test to extend over the new code in `crypt-des`, please?

besser82:
I've made some small changes to this PR to increase coverage.  Merging now.

PR #12, opened 2018-07-09
yescrypt backend

vt-alt:
This patch set implements yescrypt backend for crypt/gensalt, which is modern password hashing scheme. Original [yescrypt password hashing scheme](http://openwall.com/yescrypt/) is developed by Solar Designer (@solardiz) and based on Colin Percival's work on scrypt.

yescrypt prefix is `$y$`, but it also could `crypt()` scrypt's `$7$` hashes. Output hash size is 256-bits. `count` parameter translate to yescrypt's parameters specifying memory usage by formula `1 << (count - 1)` of MiBs, and `0` is recommended default.

First patch just verbatim copy of yescrypt sources from the current yescrypt project cvs. Second patch rename files, cut them from dead code, and modify them to work with libxcrypt. Third patch implements crypt/gensalt backend itself.

Thanks,

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -28,7 +28,8 @@ noinst_HEADERS = \
 	alg-des.h alg-hmac-sha1.h alg-md4.h alg-md5.h \
 	alg-sha1.h alg-sha256.h alg-sha512.h byteorder.h \
 	crypt-base.h crypt-obsolete.h crypt-private.h \
-	crypt-port.h test-des-cases.h
+	crypt-port.h test-des-cases.h yescrypt-sha256.h \
+	yescrypt.h yescrypt-sysendian.h
 
 lib_LTLIBRARIES = libcrypt.la
 libcrypt_la_SOURCES = \
@@ -37,7 +38,8 @@ libcrypt_la_SOURCES = \
 	crypt.c crypt-bcrypt.c crypt-des.c crypt-gensalt-static.c \
 	crypt-gensalt.c crypt-md5.c crypt-nthash.c crypt-pbkdf1-sha1.c \
 	crypt-sha256.c crypt-sha512.c crypt-static.c crypt-sunmd5.c \
-	randombytes.c
+	randombytes.c yescrypt-opt.c yescrypt-sha256.c \
+	yescrypt-common.c crypt-yescrypt.c
 
 pkgconfig_DATA = libxcrypt.pc
 # Install libcrypt.pc symlink to libxcrypt.pc file.
@@ -153,7 +155,8 @@ check_PROGRAMS = \
 	test-crypt-sunmd5 \
 	test-byteorder test-badsalt test-badsetting test-gensalt \
 	test-crypt-badargs test-short-outbuf \
-        test-getrandom-interface test-getrandom-fallbacks
+        test-getrandom-interface test-getrandom-fallbacks \
+	test-crypt-yescrypt
 
 if ENABLE_OBSOLETE_API
 libcrypt_la_SOURCES += crypt-des-obsolete.c
@@ -184,6 +187,7 @@ LOG_COMPILER = m4/skip-if-exec-format-error
 endif
 EXTRA_DIST += m4/skip-if-exec-format-error
 
+test_crypt_yescrypt_LDADD = libcrypt.la
 test_crypt_bcrypt_LDADD = libcrypt.la
 test_crypt_des_LDADD = libcrypt.la
 test_crypt_md5_LDADD = libcrypt.la
--- a/crypt-port.h
+++ b/crypt-port.h
@@ -256,4 +256,21 @@ void _xcrypt_secure_memset (void *s, size_t len)
 #define gensalt_sha_rn           _crypt_gensalt_sha_rn
 #endif
 
+#if INCLUDE_yescrypt
+#define PBKDF2_SHA256            _crypt_PBKDF2_SHA256
+#define yescrypt_encode_params_r _crypt_yescrypt_encode_params_r
+#define yescrypt_free_local      _crypt_yescrypt_free_local
+#define yescrypt_init_local      _crypt_yescrypt_init_local
+#define yescrypt_kdf             _crypt_yescrypt_kdf
+#define yescrypt_r               _crypt_yescrypt_r
+#define libcperciva_SHA256_Init  _crypt_SHA256_Init
+#define libcperciva_SHA256_Update _crypt_SHA256_Update
+#define libcperciva_SHA256_Final _crypt_SHA256_Final
+#define libcperciva_SHA256_Buf   _crypt_SHA256_Buf
+#define libcperciva_HMAC_SHA256_Init _crypt_HMAC_SHA256_Init
+#define libcperciva_HMAC_SHA256_Update _crypt_HMAC_SHA256_Update
+#define libcperciva_HMAC_SHA256_Final _crypt_HMAC_SHA256_Final
+#define libcperciva_HMAC_SHA256_Buf _crypt_HMAC_SHA256_Buf
+#endif
+
 #endif /* crypt-port.h */
--- a/crypt-yescrypt.c
+++ b/crypt-yescrypt.c
@@ -1,0 +1,109 @@
+/* Copyright (C) 2018 vt@altlinux.org
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+#include "crypt-port.h"
+#include "crypt-private.h"
+#include "byteorder.h"
+
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+#include "yescrypt.h"
+
+#if INCLUDE_yescrypt
+
+/*
+ * As OUTPUT is initialized with a failure token before gensalt_yescrypt_rn
+ * is called, in case of an error we could just set an appropriate errno
+ * and return.
+ * Since O_SIZE is guaranteed to be greater than 2, we may fill OUTPUT
+ * with a short failure token when need.
+ */
+void
+gensalt_yescrypt_rn(unsigned long count,
+                    const uint8_t *rbytes, size_t nrbytes,
+                    uint8_t *output, size_t o_size)
+{
+	/* Use one of recommended parameter sets as the 'low default'. */
+	yescrypt_params_t params = { .flags = YESCRYPT_DEFAULTS,
+		.N = 4096, .r = 32, .p = 1 };
+
+	if (count) {
+		/*
+		 * `1 << (count - 1)` is MiB usage in range of 1MiB..1GiB,
+		 * thus, count is in range of 1..11
+		 */
+		if (count <= 2) {
+			params.r = 8; /* N in 1KiB */
+			params.N = 512ULL << count;
+		} else if (count <= 11) {
+			params.r = 32; /* N in 4KiB */
+			params.N = 128ULL << count;
+		} else {
+			errno = EINVAL;
+			return;
+		}
+	}
+
+	if (!yescrypt_encode_params_r(&params, rbytes, nrbytes, output, o_size)) {
+		/*
+		 * As the output could have already been written,
+		 * overwrite it with a short failure token.
+		 */
+		output[0] = '*';
+		output[1] = '\0';
+		errno = ERANGE;
+		return;
+	}
+}
+
+void
+crypt_yescrypt_rn(const char *phrase, size_t ARG_UNUSED (phr_size),
+                  const char *setting, size_t ARG_UNUSED (set_size),
+                  uint8_t *output, size_t o_size,
+                  ARG_UNUSED(void *scratch), ARG_UNUSED(size_t s_size))
+{
+	yescrypt_local_t local;
+	uint8_t *retval;
+
+	if (o_size < 3) {
+		errno = ERANGE;
+		return;
+	}
+	if (yescrypt_init_local(&local)) {
+		errno = ENOMEM;
+		return;
+	}
+	retval = yescrypt_r(NULL, &local,
+	    (const uint8_t *)phrase, strlen(phrase),
+	    (const uint8_t *)setting, NULL,
+	    output, o_size);
+	if (yescrypt_free_local(&local) ||
+	    !retval) {
+		/*
+		 * As the output could have already been written,
+		 * overwrite it with a failure token.
+		 */
+		output[0] = '*';
+		output[1] = '0';
+		output[2] = '\0';
+		errno = EINVAL;
+	}
+}
+
+#endif /* INCLUDE_yescrypt */
--- a/hashes.lst
+++ b/hashes.lst
@@ -28,6 +28,8 @@
 # DES) must be last.
 #
 #name    gs_suffix  h_prefix  nrbytes  flags
+yescrypt :          $y$       16       STRONG
+yescrypt :          $7$       16       STRONG
 bcrypt   _b         $2b$      16       STRONG,FREEBSD,NETBSD,OPENBSD,SOLARIS,DEFAULT
 bcrypt   _a         $2a$      16       STRONG,FREEBSD,NETBSD,OPENBSD,SOLARIS
 bcrypt   _x         $2x$      16       STRONG,FREEBSD,NETBSD,OPENBSD,SOLARIS
--- a/test-crypt-yescrypt.c
+++ b/test-crypt-yescrypt.c
@@ -1,0 +1,73 @@
+/* Copyright (C) 2018 vt@altlinux.org
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+#include "crypt-port.h"
+#include "crypt-base.h"
+
+#include <stdio.h>
+
+#if INCLUDE_yescrypt
+
+static int
+test(const char *p, const char *s, const char *m)
+{
+	char *cp;
+	struct crypt_data output;
+	cp = crypt_r(p, s, &output);
+	if (!cp)
+		return 1;
+	if (strcmp(m, cp)) {
+		fprintf(stderr, "ERROR: %s %s -> %s\n\t(expected %s)\n",
+		    p, s, cp, m);
+		return 1;
+	} else {
+		fprintf(stderr, "   ok: %s %s -> %s\n", p, s, cp);
+		return 0;
+	}
+}
+
+int
+main (void)
+{
+	int result = 0;
+
+	result |= test("pleaseletmein", "$7$C6..../....SodiumChloride",
+	    "$7$C6..../....SodiumChloride$kBGj9fHznVYFQMEn/qDCfrDevf9YDtcDdKvEqHJLV8D");
+	result |= test("pleaseletmein", "$7$06..../....SodiumChloride",
+	    "$7$06..../....SodiumChloride$ENlyo6fGw4PCcDBOFepfSZjFUnVatHzCcW55.ZGz3B0");
+	result |= test("pleaseletmein", "$y$jD5.7$LdJMENpBABJJ3hIHjB1Bi.",
+	    "$y$jD5.7$LdJMENpBABJJ3hIHjB1Bi.$HboGM6qPrsK.StKYGt6KErmUYtioHreJd98oIugoNB6");
+
+	result |= test("", "$y$jD5.7$",  "$y$jD5.7$$JD8dsR.nt1ty0ltQ2HHwauaDRoOUIEaA5i.vpj2nyL.");
+	result |= test("", "$y$jD5.7$$", "$y$jD5.7$$JD8dsR.nt1ty0ltQ2HHwauaDRoOUIEaA5i.vpj2nyL.");
+	result |= test("test", "$y$", "*0");
+	result |= test("test", "*0", "*1");
+	result |= test("test", "*1", "*0");
+
+	return result;
+}
+
+#else
+
+int
+main (void)
+{
+  return 77; /* UNSUPPORTED */
+}
+
+#endif /* INCLUDE_yescrypt */
--- a/test-gensalt.c
+++ b/test-gensalt.c
@@ -216,6 +216,14 @@ static const char *const bcrypt_y_expected_output[] =
   "$2y$05$mAyzaIeJu41dWUkxEbn8hO"
 };
 #endif
+#if INCLUDE_yescrypt
+static const char *yescrypt_expected_output[] = {
+  "$y$j9T$MJHnaAkegEVYHsFKkmfzJ1",
+  "$y$j9T$PKXc3hCOSyMqdaEQArI62/",
+  "$y$j9T$ZAFlICwYRETzIzIjEIC86.",
+  "$y$j9T$UqGBkVu01rurVZqgNchTB0"
+};
+#endif
 
 struct testcase
 {
@@ -294,6 +302,9 @@ static const struct testcase testcases[] =
   { "$2a$",  bcrypt_a_expected_output, 29,  0, 0 },
   { "$2x$",  bcrypt_x_expected_output, 29,  0, 0 },
   { "$2y$",  bcrypt_y_expected_output, 29,  0, 0 },
+#endif
+#if INCLUDE_yescrypt
+  { "$y$",   yescrypt_expected_output, 29, 29, 0 },
 #endif
   { 0, 0, 0, 0, 0 }
 };
--- a/yescrypt-opt.c
+++ b/yescrypt-opt.c
@@ -28,12 +28,20 @@
  * online backup system.
  */
 
+#include "crypt-port.h"
+
+#pragma GCC diagnostic ignored "-Wcast-align"
+#ifdef __clang__
+#pragma GCC diagnostic ignored "-Wtautological-constant-out-of-range-compare"
+#endif
+
 /*
  * AVX and especially XOP speed up Salsa20 a lot, but this mostly matters for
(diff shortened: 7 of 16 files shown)


besser82:
First of all, thank you for your contribution! @ldv-alt already announced this PR was incoming.

Before I start with a detailed review, there are many build failures on TravisCI - especially with Clang - caused by `-Werror=cast-align`.  Let's sort these out first.

Coverage is also very poor:
```
Summary coverage rate:
  lines......: 83.5% (2602 of 3116 lines)
  functions..: 92.3% (144 of 156 functions)
  branches...: no data found
```

Makefile.am
```diff
@@ -127,6 +129,7 @@ check_PROGRAMS = \
 	test-crypt-pbkdf1-sha1 test-crypt-sha256 test-crypt-sha512 \
 	test-crypt-sunmd5 \
 	test-byteorder test-badsalt test-gensalt test-short-outbuf
+	test-crypt-yescrypt
```
besser82: You have missed a trailing blackslash `\` on the previous line.

yescrypt-sysendian.h
```diff
@@ -0,0 +1,182 @@
+/*-
```
besser82: This file almost duplicates the contents of `byteorder.h`.  The functions for 16 bit types could be added there and then yescrypt files can be modified to use them.

yescrypt-sha256.c
```diff
@@ -0,0 +1,648 @@
+/*-
```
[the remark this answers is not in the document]
besser82: Implementing HMAC-SHA256 is pretty simple as well.  See `alg-hmac-sha1.c` for a reference.

yescrypt-sha256.h
```diff
@@ -0,0 +1,121 @@
+/*-
```
besser82: This is duplicate code. The SHA256 hash algorythm is already implemented in `alg-sha256.h`. Implementing HMAC-SHA256 is pretty simple as well. See `alg-hmac-sha1.h` for a reference.

vt-alt: It's slightly different in API (like have additional `tmp32` parameter), and implements HMAC_SHA256 on top of this, and PBKDF2_SHA256 on top of that (having calls to things like SHA256_Pad_Almost(), which `alg-sha256.c` does not provide).

yescrypt-sysendian.h
besser82: From a quick check in coverage report, it seems the 16 bit conversions are not used anyways.  You should drop this file and replace it usage with `byteorder.h` as it can be done in a really safe way.

yescrypt-opt.c
```diff
@@ -0,0 +1,1521 @@
...
+ * need a 4-byte displacement.
+ */
+#define PWXFORM_WRITE_OFFSET 0x7c
+
+#define PWXFORM_WRITE \
+	WRITE_X(*(salsa20_blk_t *)(Sw - PWXFORM_WRITE_OFFSET)) \
+	Sw += 64;
+
+#define PWXFORM { \
+	uint8_t *Sw = S2 + w + PWXFORM_WRITE_OFFSET; \
```
besser82: This one must be changed to `uint8_t alignas (salsa20_blk_t *) …`.

yescrypt-sha256.h
besser82: `SHA256_Pad_Almost()` isn't used in yescrypt code-base anywhere; it's a simple helper function, that is called by this specific implementation of `PBKDF2_SHA256()`.

Implementing this functionality in several seperate reusable files (and adding the official RFC and FIPS-180 test-vectors) shouldn't be to much of a problem.

crypt-yescrypt.c
```diff
@@ -0,0 +1,108 @@
...
+			params.r = 32; /* N in 4KiB */
+			params.N = 128ULL << count;
+		} else {
+			errno = EINVAL;
+			return;
+		}
+	}
+
+	if (!yescrypt_encode_params_r(&params, rbytes, nrbytes, output, o_size)) {
+		/*
```
besser82: `make_failure_token()` from `crypt.c` could be used here.  It just needs some small tweaks to the general code-base.

test-gensalt.c
```diff
@@ -98,6 +98,14 @@ static const char *const bcrypt_y_expected_output[] =
   "$2y$05$mAyzaIeJu41dWUkxEbn8hO"
 };
 #endif
+#if INCLUDE_yescrypt
```
besser82: What about testing gensalt for scrypt `$7$` here as well?

crypt-yescrypt.c
besser82: Another (and the better idea instead of my first sugestion) would be to allocate a temporary buffer with `o_size` but not more than max. needed length and strncpy it to output after success.

crypt-yescrypt.c
```diff
@@ -0,0 +1,108 @@
...
+		errno = ENOMEM;
+		return;
+	}
+	retval = yescrypt_r(NULL, &local,
+	    (const uint8_t *)phrase, strlen(phrase),
+	    (const uint8_t *)setting, NULL,
+	    output, o_size);
+	if (yescrypt_free_local(&local) ||
+	    !retval) {
+		/*
```
besser82: Same goes here.  (temporary buffer to keep error token and such)

yescrypt-opt.c
solardiz: No. And out of those you list, I've already tested the code on PPC64 and ARM. Sw is deliberately unaligned, which is then corrected with inclusion of PWXFORM_WRITE_OFFSET also in the actual writes, keeping them aligned. We could exclude this logic on archs where it's not known to help (it is on x86[-64]), but I'd rather not complicate the code with conditional compilation unnecessarily.

solardiz:
I am very uncomfortable making or even approving changes to yescrypt itself for this merge to happen. I strongly prefer that we keep yescrypt itself as-is. I intend to continue to maintain it upstream, and I really don't want to create extra work and risk of bugs (now and for further merges) out of thin air.

I understand that it's not great to have e.g. duplicate SHA-256 code. Maybe we can tackle this the other way around - change the rest of libxcrypt to use the SHA-256, HMAC, PBKDF2 code provided along with yescrypt? It's code I ran tests on and know it works correctly and is optimized. What's currently in libxcrypt is of unknown status from my perspective. ;-)

As to making yescrypt the default: yes, we'll probably need to move away from bcrypt in the next few years because of advances in FPGA and GPU attacks. Those you linked to aren't even the best results available - we link to our better results for bcrypt on FPGA right off the yescrypt homepage. ;-) However, yescrypt with its current parameters relies on availability of SSE2 or the like for decent performance. SSE2 is implied on x86-64, but not on 32-bit x86. I already mentioned to @ldv-alt that maybe libxcrypt should include building of an SSE2 version of the library on 32-bit x86 by default, letting the dynamic linker load it where SSE2 is supported (that is, on Pentium 4 / Athlon or newer). Speaking of non-x86 archs, I also intend to add NEON/ASIMD support (again, implied on Aarch64, but optional on 32-bit ARM), but am unsure whether it's worth adding anything else. Maybe it's OK'ish that some of those will run the slower scalar code (especially the 64-bit register-rich ones like PPC64 are OK'ish running it).

yescrypt-sha256.h
solardiz: `SHA256_Pad_Almost()` _is_ indirectly used in yescrypt, and I added it specifically for its use in yescrypt. This happens via `PBKDF2_SHA256()` when it's called by yescrypt with c=1, as yescrypt always does (and as scrypt always did). This is optimization for this special case of c=1. While we could simplify things for libxcrypt in particular and lose maybe just ~1% of performance on this (since yescrypt spends most time in its own heavy processing and not in its initial/final PBKDF2 steps), I am uncomfortable maintaining two versions of the code and for some larger deployments a combination of micro-optimizations like this was actually needed to squeeze out the last few missing percent of performance to meet server request rate capacity requirements. An option for libxcrypt is to revert to using Colin's sha256.c from the scrypt tree as-is, though. I maintained compatibility with it.

yescrypt-sysendian.h
solardiz: I'm OK dropping the unused functions (inherited from scrypt, which in turn shared them with code in FreeBSD - which is why). That's a simple enough change for further merges (from scrypt). I'm not OK with more invasive changes.

yescrypt-sha256.c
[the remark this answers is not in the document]
solardiz: How about we make this the new alg-sha256.c?

solardiz:
I forgot to say: I am OK making changes to @vt-alt's wrappers for integration of yescrypt. Those are specific to the integration into libxcrypt, so are subject to change for the merge to happen.

zackw:
Quickly from my phone: I endorse the idea of replacing the existing sha256
implementation with the one from yescrypt. I was thinking of suggesting
that myself, just because it would reduce the amount of LGPL code — I have
a long-term daydream of getting the BSDs to pick up this library, which
ain’t gonna happen as long as it isn’t permissively licensed. Also, the
code was originally written by Ulrich Drepper, who believed in “if it was
hard to write it should be hard to understand,” ugh.

ldv-alt:
On Tue, Jul 10, 2018 at 05:07:08AM -0700, Solar Designer wrote:
> I already mentioned to @ldv-alt that maybe libxcrypt should include building of an SSE2 version of the library on 32-bit x86 by default, letting the dynamic linker load it where SSE2 is supported (that is, on Pentium 4 / Athlon or newer).

I think of adding a configure check whether __SSE2__ is defined on x86,
and installing an extra SSE2 version of the library into $(libdir)/sse2
when __SSE2__ is not defined on x86.

zackw:
@solardiz 
> Semi-off-topic for this issue, but I'd also like to offer my permissively licensed MD4 and MD5. IIRC, right now libxcrypt uses Ulrich's MD5 and a third-party hack of my MD5 turned into MD4 (whereas I directly have MD4 as well). I'm not convinced it's a good idea to even have MD4 in here (for NT hashes perhaps?), but if they do I think my original MD4 is better than almost the same thing via a third-party.

If you've got time to make a PR, I'll merge it. :smile: 

Yes, MD4 is present only to support NTHASH ($3$ format).  I'm not sure why @besser82 wanted to include that in the first place, but I don't think there's any danger of any new hashes adopting it either.

zackw:
@besser82 
> @zackw As soon as this is ready to merge: What about changing the default hashing algorythm to yescrypt? bcrypt seams to become weaker.

Sounds good to me!  It might even make sense to demote glibc's SHA256 and SHA512 to "too weak for new passwords" at that point.

yescrypt-opt.c
solardiz: "Fixing" the PWXFORM_WRITE_OFFSET thing to mute the warning e.g. by using `salsa20_blk_t *` pointers results in much worse code. It's just very helpful to the compiler to show that byte offsets can be maintained throughout the code instead of having to do left/right shifts by 6 or whatever. Also, this pattern and warning is seen not only here, but also in many other places in yescrypt-opt.c - e.g. yescrypt pwxform is deliberately designed such that the S-box byte offsets rather than element indices are obtained by ANDing a value with a mask and not needing any shifts. So I think we simply need to mute the warning in the compiler. That warning and avoiding such patterns can be helpful in other and non-performance-critical code, but not here.

hashes.lst
```diff
@@ -25,6 +25,8 @@
 # DES) must be last.
 #
 #name    gs_suffix  h_prefix  nrbytes  flags
+yescrypt :          $y$       15       STRONG
+yescrypt :          $7$       15       STRONG
```
solardiz: Why 15? Shouldn't these be 16? Somehow they're 15 for SHA-crypt as well. I guess it's to more optimally use the base-64 chars, arriving at a whole number of them. I don't really mind, but is it common and would everyone be comfortable with it? I imagine some compliance guidelines could require 128-bit or larger salts, and then it'd be a pity that a system uses 120-bit ones instead.

yescrypt-sysendian.h
solardiz: Dropped the unused `sysendian.h` functions in my yescrypt tree.

crypt-yescrypt.c
vt-alt: Why mess with allocation and copying, if I just can write 2-3 char? My choice is better in terms of performance and simplicity of code. I also don't need token alternation functionality from `make_failure_token`. You also have read out of bounds in `make_failure_token`.

test-gensalt.c
vt-alt: yescrypt code does not generate settings for (obsolete?) `$7$` prefix. Only crypt could encode already existing `$7$` hashes.

zackw:
I am confused about the state of this PR.  Is it worthwhile for me to do a detailed review?  If it is, I can try to find time to do that in the next few days, but it seems like enough changes have already been requested that I should wait for the next round,

I am concerned that y'all seem to be arguing over cycle-count efficiency of the cryptographic primitives.  I don't think it makes sense to worry about that, because the higher level is going to iterate the primitives until the overall operation is _slow_ enough, anyway.  We need to be _reasonably_ close to optimum so that end-user choice of round count isn't wrong by orders of magnitude (so, for instance, I am not going to ask for the total removal of the assembly optimizations) but we do not need to squeeze cycles the way we might if we were writing a TLS library.

I am also concerned that, in the name of efficiency, the new code is pulling type shenanigans that may cause compilers to declare its behavior undefined, and may also make it harder for humans to understand its memory access behavior.  I want the code in this library to be _obviously correct_ first, and efficient second.

solardiz:
I expect that @vt-alt will update this PR, but not in a major way (e.g., I've since avoided a few gcc warnings beyond -Wall in my tree, which should be merged in here).

I agree that for libxcrypt cleaner code would be preferable over squeezing out every last bit of performance. However, there are users of yescrypt where those last bits of performance are needed (large deployments where this translates into meeting or not the performance SLAs on multi-server yescrypt password hashing clusters - and yes, some of these already exist), and I'm not eager about maintaining very different code versions (this is a higher risk of bugs in at least one of them, and less testing for each one of them - not to mention more work).

I've been very careful to avoid instances of undefined behavior, except for the widely-accepted addressing of different types via a union (if a compiler breaks this, a lot of software out there will break, so it's a de-facto standard way to deal with aliasing). All aliasing going on in yescrypt is between types that are members of a union.

For easier understanding, there's `yescrypt-ref.c` in my tree - the purposefully not optimized reference implementation. Instead of aliasing, it accesses everything via `uint32_t`. We do not recommend it for actual use as it's many times slower (not just a few percent). In yescrypt 1.0.x, there's also an in-between implementation: optimized, but without SIMD. It has aliasing between `uint32_t` and `uint64_t` via a union, and endianness-converting copying where needed. I plan on dropping that one for yescrypt 1.1+, and in fact this PR already uses the combined optimized scalar + SIMD source file (thus, with aliasing between 32-, 64-, and 128-bit values via the union) that will be the basis for yescrypt 1.1+'s. If you insist, we could revert to using yescrypt 1.0.3's `yescrypt-opt.c`, which lacks SIMD support, but that makes it 2x slower on x86_64 and 4x+ slower on 32-bit x86.

Maybe we should refer to the availability of `yescrypt-ref.c` from somewhere in libxcrypt tree for those who want to understand yescrypt's memory access behavior or whatever through simpler code?

I think that at some point (hopefully soon) we'll need to introduce runtime self-test similar to what I implemented for bcrypt. This will ensure yescrypt will refuse to run if miscompiled and will also help it effectively zeroize its buffers (better than with explicit zeroization). But if we try to do that right away, we'll get scope creep and "never" merge anything.

vt-alt:
I plan to rebase again against moving head, and apply minor changes from Solar and what is discussed reasonable about my part of code. I will *not* change yescrypt code (except disabling warnings and making it compile on clang and removing dead code). If you want (me?) to merge alg-sha256 with yescrypt-sha256 probably that could be done after this PR? (Also I plan to make different PR on the base of this one).

solardiz:
The alg-sha256 merge should definitely be separate from this PR (or its replacement). One issue to discuss there is the zeroization of SHA-256 context buffers, which would be too slow to do in `SHA256_Final()` when invoked from iterated hashes such as `sha256crypt`, so I guess a lower-level primitive (which we have) would need to be available and used from there (just like our PBKDF2 code uses, for the same reason), and the zeroization only done after all the iterations (not per iteration).

vt-alt:
Also, we are already using this modifications in ALT so we are quite interested in pushing it upstream.

zackw:
@solardiz wrote:
> I agree that for libxcrypt cleaner code would be preferable over squeezing out every last bit of performance. However, there are users of yescrypt where those last bits of performance are needed (large deployments where this translates into meeting or not the performance SLAs on multi-server yescrypt password hashing clusters - and yes, some of these already exist), and I'm not eager about maintaining very different code versions

I definitely understand and agree with a goal of minimizing code divergence.

> I've been very careful to avoid instances of undefined behavior, except for the widely-accepted addressing of different types via a union

Good to hear.

> ... If you insist, we could revert to using yescrypt 1.0.3's yescrypt-opt.c, which lacks SIMD support, but that makes it 2x slower on x86_64 and 4x+ slower on 32-bit x86.

2x is enough of a difference that I would rather keep the faster code.  It _is_ a problem if someone on lower-end hardware (underpowered Android phones are probably the most important class of "lower-end hardware" nowadays) has to use a round count that's so low that an adversary can still brute-force it.

> Maybe we should refer to the availability of yescrypt-ref.c from somewhere in libxcrypt tree for those who want to understand yescrypt's memory access behavior or whatever through simpler code?

That sounds like a good idea in general.  When I said "memory access behavior", though, I meant things like "can I be _sure_ that this function does not read or write memory it's not entitled to read or write?" and for that, a reference implementation doesn't help.

> I think that at some point (hopefully soon) we'll need to introduce runtime self-test similar to what I implemented for bcrypt. This will ensure yescrypt will refuse to run if miscompiled and will also help it effectively zeroize its buffers (better than with explicit zeroization). But if we try to do that right away, we'll get scope creep and "never" merge anything.

I'm planning to do that at the crypt.c level for _all_ hash algorithms, as a replacement for the current use of swapcontext.  But I can't commit to any timeframe on that, and indeed it should not block this merge.

solardiz:
Reference implementation does partially help address the issue of out-of-bounds access: if this different and simpler code computes the same hash values, then it's more likely the optimized code doesn't do out-of-bounds accesses either. Ditto for testing on other platforms. For further assurance, albeit also only on specific tested inputs, there's ASan.

Runtime self-test for all hash algorithms would be great to have, but it's also tricky to do optimally as we'd need to be requesting iteration counts below the algorithms' publicly exported minimums. For example, SHA-crypt has a minimum of 1000, but for self-test we'd invoke e.g. 7, 21, or 42 (IIRC, one of these is the lowest to explore all code paths - need to find out), so we'd need to somehow export this functionality only for the library's own self-test, but not for its users. We could also record successful self-test status in a thread-local variable and then skip self-test for similar setting strings on further invocations.

besser82:
@vt-alt Any news on this?

vt-alt:
@besser82 I will try to push rebased version with all small fixes before Monday.

vt-alt:
Excuse ma for delay, I'm on vacation until 20 Aug, then I will continue with this nice pull request.

vt-alt:
I pushed my commits rebased over the new head, containing small fixes including Solar's. Travis seem to build it OK, even though error is reported it's the same as with two latest head commits, due to Travis' build env being broken.

besser82:
Please rebase again on recent develop branch.  TravisCI should work then.

ldv-alt:
The change of recommended password hashing scheme in crypt.5 without a corresponding change of the default prefix doesn't look consistent.  I suggest moving the change of defaults and recommendations to a separate follow-up commit.

besser82:
@ldv-alt LGTM, so far.

@zackw Would you mind to review this?

crypt-port.h
```diff
@@ -256,4 +256,22 @@ void _xcrypt_secure_memset (void *s, size_t len)
 #define gensalt_sha_rn           _crypt_gensalt_sha_rn
 #endif
 
+#if INCLUDE_yescrypt
+#define PBKDF2_SHA256            _crypt_PBKDF2_SHA256
+#define yescrypt_digest_shared   _crypt_yescrypt_digest_shared
```
solardiz: If dropping ROM support, it's also fairly easy and non-invasive to drop `yescrypt_digest_shared`.

yescrypt-opt.c
```diff
@@ -0,0 +1,1451 @@
...
+	    passwd, passwdlen, salt, saltlen,
+	    flags, N, r, p, t, NROM, buf, buflen);
+#ifndef SKIP_MEMZERO
+	if (passwd == dk)
+		insecure_memzero(dk, sizeof(dk));
+#endif
+	return retval;
+}
+
+yescrypt_binary_t *yescrypt_digest_shared(yescrypt_shared_t *shared)
```
solardiz: If dropping ROM support, it's also fairly easy and non-invasive to drop yescrypt_digest_shared.

yescrypt-opt.c
```diff
@@ -0,0 +1,1451 @@
...
+	tag = (uint64_t *)
+	    ((uint8_t *)shared->aligned + shared->aligned_size - 48);
+
+	if (tag[0] != YESCRYPT_ROM_TAG1 || tag[1] != YESCRYPT_ROM_TAG2)
+		return NULL;
+
+	le64enc(digest.uc, tag[2]);
+	le64enc(digest.uc + 8, tag[3]);
+	le64enc(digest.uc + 16, tag[4]);
+	le64enc(digest.uc + 24, tag[5]);
```
solardiz: I think these are the only uses of `le64enc()`, so they can go as well.

yescrypt-sysendian.h
```diff
@@ -0,0 +1,115 @@
...
+#pragma GCC diagnostic ignored "-Wtype-limits"
+#pragma GCC diagnostic ignored "-Wcast-qual"
+
+/* Avoid namespace collisions with BSD <sys/endian.h>. */
+#define be32dec libcperciva_be32dec
+#define be32enc libcperciva_be32enc
+#define be64enc libcperciva_be64enc
+#define le32dec libcperciva_le32dec
+#define le32enc libcperciva_le32enc
+#define le64enc libcperciva_le64enc
```
solardiz: Can drop `le64enc()`.

yescrypt-sysendian.h
```diff
@@ -0,0 +1,115 @@
...
+
+	p[0] = x & 0xff;
+	p[1] = (x >> 8) & 0xff;
+	p[2] = (x >> 16) & 0xff;
+	p[3] = (x >> 24) & 0xff;
+	p[4] = (x >> 32) & 0xff;
+	p[5] = (x >> 40) & 0xff;
+	p[6] = (x >> 48) & 0xff;
+	p[7] = (x >> 56) & 0xff;
+}
```
solardiz: Can drop le64enc().

yescrypt.h
```diff
@@ -0,0 +1,257 @@
...
+ */
+typedef yescrypt_region_t yescrypt_shared_t;
+typedef yescrypt_region_t yescrypt_local_t;
+
+/**
+ * Two 64-bit tags placed 48 bytes to the end of a ROM in host byte endianness
+ * (and followed by 32 bytes of the ROM digest).
+ */
+#define YESCRYPT_ROM_TAG1 0x7470797263736579 /* "yescrypt" */
+#define YESCRYPT_ROM_TAG2 0x687361684d4f522d /* "-ROMhash" */
```
solardiz: Can drop the YESCRYPT_ROM_TAG* defines.

hashes.lst
solardiz: As discussed, these 15's should be changed to 16's.

vt-alt:
@ldv OK, already reverted that in last push.

yescrypt.h
vt-alt: Can't, since they are used in yescrypt_kdf_body in conditional which is not dropped yet.

vt-alt: I feared to remove these conditionals in case they are supposed to be called with some settings.

vt-alt: @solardiz Maybe you could add commit later with proper removal of such code?

vt-alt:
Just applied all changes that @solardiz suggested and reminded of.

yescrypt.h
solardiz: OK, let's keep these for now. I agree it's less invasive that way. Thanks.

hashes.lst
```diff
@@ -28,6 +28,8 @@
 # DES) must be last.
 #
 #name    gs_suffix  h_prefix  nrbytes  flags
+yescrypt :          $y$       16       STRONG
+yescrypt :          $7$       16       STRONG
```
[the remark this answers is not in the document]
solardiz: No, it's not in any OS I'm aware of. However, it's in libsodium (through libsodium's use of a fork of my enhanced scrypt tree that later led to yescrypt), which is a popular library and is available via various language bindings. So such hashes may be in use, but probably not yet by an OS. Since scrypt didn't originally come with an official crypt(3)-style encoding, there are many other encodings also in use. John the Ripper -jumbo includes support for `$7$`, as well as some others.

`$7$` was the encoding I had suggested for scrypt in 2012 or so, and it was in an informal specification document maintained by the author of the eventual scrypt RFC, but ultimately the RFC does not include any crypt(3)-style encoding.

Classic scrypt hashes may also be encoded (and with fewer characters) under the `$y$` prefix (where the variable length integer encoded right after that prefix is the flavor, with single `.` character meaning 0, which is classic scrypt). This is the encoding that `yescrypt_encode_params()` will use if invoked with a params structure with flags=0, requesting classic scrypt.

So for the current yescrypt tree `$7$` is a decoding-only compatibility prefix.

test-crypt-yescrypt.c
```diff
@@ -0,0 +1,73 @@
...
+	result |= test("pleaseletmein", "$7$06..../....SodiumChloride",
+	    "$7$06..../....SodiumChloride$ENlyo6fGw4PCcDBOFepfSZjFUnVatHzCcW55.ZGz3B0");
+	result |= test("pleaseletmein", "$y$jD5.7$LdJMENpBABJJ3hIHjB1Bi.",
+	    "$y$jD5.7$LdJMENpBABJJ3hIHjB1Bi.$HboGM6qPrsK.StKYGt6KErmUYtioHreJd98oIugoNB6");
+
+	result |= test("", "$y$jD5.7$",  "$y$jD5.7$$JD8dsR.nt1ty0ltQ2HHwauaDRoOUIEaA5i.vpj2nyL.");
+	result |= test("", "$y$jD5.7$$", "$y$jD5.7$$JD8dsR.nt1ty0ltQ2HHwauaDRoOUIEaA5i.vpj2nyL.");
+	result |= test("test", "$y$", "*0");
+	result |= test("test", "*0", "*1");
+	result |= test("test", "*1", "*0");
```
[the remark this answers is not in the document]
solardiz: I think some of the uncovered code is not reachable (the remnants of ROM support beyond checking that there must be no ROM), and some is (classic scrypt). We should probably add classic scrypt test vectors using both `$y$.` and `$7$` prefixes. Some test vectors may be taken from libsodium.

crypt-yescrypt.c
```diff
@@ -0,0 +1,109 @@
...
+	uint8_t *retval;
+
+	if (o_size < 3) {
+		errno = ERANGE;
+		return;
+	}
+	if (yescrypt_init_local(&local)) {
+		errno = ENOMEM;
+		return;
+	}
```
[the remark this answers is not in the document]
solardiz: I agree this should "just return".

In further maintenance of yescrypt, I'm willing to take responsibility to set errno (or preserve errno set by a deeper failing call) when any of yescrypt's exported functions fails.

crypt-yescrypt.c
```diff
@@ -0,0 +1,109 @@
...
+			params.N = 512ULL << count;
+		} else if (count <= 11) {
+			params.r = 32; /* N in 4KiB */
+			params.N = 128ULL << count;
+		} else {
+			errno = EINVAL;
+			return;
+		}
+	}
+
```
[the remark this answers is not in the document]
solardiz: `N` (block count) is in fact specified in units of `r` (block size). In the yescrypt documentation, I recommend `r=8` for total sizes of 2 MiB and less, and `r=32` above that. This has to do with typical per-core last-level cache sizes of current CPUs.

solardiz:
> The largest thing I would like to ask for is, I think we should go ahead and support $7$ (scrypt) in gensalt. We do that for all the other compatibility hashes, including the ones that really shouldn't be used anymore.

Maybe that's a reason to reconsider and stop doing it for hashes "that really shouldn't be used anymore"?

The current yescrypt tree does support encoding of salts for classic scrypt hashes, but using the `$y$.` prefix. It also supports decoding of `$7$`, and I'll consider adding a few more also for decoding only (unfortunately, there are many different ones in use for scrypt hashes - e.g., Cisco's `$9$`). But generating them all?.. I didn't plan on supporting that.

besser82:
@zackw I'll merge this PR as I have already a local branch which addresses the changes / fixes you mentioned / requested during your review.

I will push and file my branch with the fixes for review during later today.  I dont't want the current PR to become a never ending story…

@vt-alt Thanks for your initial work!

vt-alt:
@zackw Thanks for the review and @besser82 for applying fixes in the new pull request. I would like to add some comments, though.

crypt-yescrypt.c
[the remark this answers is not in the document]
vt-alt: I think proposed comment section is turned out not enough explanatory, and I proposed change in the new pull request, there https://github.com/besser82/libxcrypt/pull/27#pullrequestreview-149541425 I post here to bring attention of people subscribed to this pull who may be interested in that change.

crypt-yescrypt.c
```diff
@@ -0,0 +1,109 @@
...
+	    !retval) {
+		/*
+		 * As the output could have already been written,
+		 * overwrite it with a failure token.
+		 */
+		output[0] = '*';
+		output[1] = '0';
+		output[2] = '\0';
+		errno = EINVAL;
+	}
```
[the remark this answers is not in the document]
vt-alt: > This has two problems. Most important, you must not simply set output to *0. You must check whether setting is already *0, and if it is, you must instead set output to *1.

I disagree. This alternation logic is _already performed_ by `crypt_rn`, and this function is not user visible API, and is never called with `*`, `*0` or `*1` setting.

> It's probably not possible to reach this point with setting equal to *0,

It's certainly not possible.

> but "a failure token will never compare equal to the setting string passed to crypt()" is a security-critical guarantee made by the API contract

And this is not implementing `crypt` API. This is internal function.

> so it is worth including defensive code to ensure it

With a price of not reachable, not covered dead code.

> even if we think it can't ever happen.

We don't "think" it can't ever happen, it's fact that it will never happen, it's coded that way, that low level crypt backends never need to implement that part of API.

hashes.lst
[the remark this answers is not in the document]
vt-alt: Maybe we should not implement `$7$` prefix at all in libcxrypt?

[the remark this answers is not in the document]
solardiz: Not implementing the `$7$` prefix in libxcrypt at all (at least for now, until/unless we see any demand for it) is OK with me.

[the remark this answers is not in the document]
vt-alt: Stating an argument to not add `$7$` to linxcrypt: we are (in ALT) never intended to add it, it is just side effect of what yescrypt supports, to implement something that is not needed just for the sake symmetry or possibility is strange; it is purely virtual; and scrypt is already properly supported under `$y$` prefix.

PR #14, opened 2018-07-14
Replace crypt-sha{256,512}.c with a BSD-licensed implementation.

besser82:
In fact the change boils down to a change in the license boilerplate, since the code used in glibc was taken from Ulrich Drepper's implementation, which was given to Public Domain by him.


besser82:
@zackw Could you have another look, please?  If ok, I'll merge on CLI, then.

Issue #15, opened 2018-07-20
Segfault when pass is NULL and when generating NTHashes


sedrubal:
Following (invalid) code will produce a segfault, because `phrase` is `NULL`:

```c
#include <crypt.h>
#include <stdio.h>

int main(int argc, const char *argv[])
{
        struct crypt_data data;
        data.initialized = 0;
        const char* hash = crypt_r((char*) NULL, "$3$$", &data);
        printf("%s\n", hash);
        return 0;
}
```

I'm not sure if this is intended. Other hashes return "*0", when `pass` is `NULL`.

zackw:
Clearly the behavior should not depend on which hashing method is in use.

There is an argument that passing NULL as `phrase` should return `*0` for all hashes, but there is also an argument that it should _crash_ for all hashes.  (See https://blogs.msdn.microsoft.com/oldnewthing/20060927-07/?p=29563 and also consider what happens if you pass a pointer to five `'a'`s followed by unmapped address space as `phrase`; we _can't_ reliably do anything other than crash in that case.)

@besser82 which seems more appropriate to you?

besser82:
Passing in `NULL` as `phrase` should return `*0` and set `errno = EINVAL` with all hashes, since passing a pointer to `NULL` is a valid parameter by the C-standard; checking for a `NULL` pointer is valid and doesn't cause random side-effects.  Thus we should signal an error in that case.

zackw:
Which bit of the C standard are you thinking of?  The general rule is that passing NULL to a C library function provokes undefined behavior (7.1.4p1, very first sentence):

> Each of the following statements applies unless explicitly stated otherwise in the detailed
descriptions that follow: If an argument to a function has an invalid value (such as a value
outside the domain of the function, or a pointer outside the address space of the program,
or a null pointer, or a pointer to non-modifiable storage when the corresponding
parameter is not const-qualified) or a type (after promotion) not expected by a function
with variable number of arguments, the behavior is undefined.

`crypt` itself isn't in C, but it _is_ in POSIX, so that spec could "explicitly state otherwise", but it doesn't: compare the language about the _key_ argument in  http://pubs.opengroup.org/onlinepubs/9699919799/functions/crypt.html to the language about the _ptr_ argument in http://pubs.opengroup.org/onlinepubs/9699919799/functions/free.html .

besser82:
[...]

PR #17, opened 2018-07-24
Issue #15: predictable behavior when crypt args are NULL or invalid.

zackw:
This patch introduces the following semantics for NULL or invalid
`phrase` and `setting` arguments to the `crypt*` functions:

 - If `setting` is NULL, all of the functions will fail (in their
   usual manner) with errno set to EINVAL, regardless of what `phrase`
   is.  (For crypt_r/rn/ra, problems with the `data` argument may be
   detected first.)

 - If `setting` is a valid string (in the sense that calling `strlen`
   on it will not crash the program), but `phrase` is NULL, again all
   of the functions will fail with errno set to EINVAL.

 - If both `setting` and `phrase` are non-NULL, and one of them is an
   _invalid_ string (in the sense that calling `strlen` on it _will_
   crash the program), then the program will crash.  The crash is
   guaranteed to happen before any code from a hashing method is executed.

 - If `phrase` is NULL and `setting` is an invalid string, then it is
   unspecified whether the function will fail or whether the program
   will crash, but it will be one of those two things, and, again, no
   code from a hashing method will be executed.  (The actual behavior
   right now is that the function will fail *unless* `setting` is
   non-NULL and there are fewer than two readable bytes at that memory
   location, but I don't want to make that a contract.)

I would somewhat prefer the simpler semantics of always crashing on
NULL as well as on invalid strings, but there was some sentiment in
favor of returning failure for NULL, so I've done this to the maximum
extent that is practical.  The complicated behavior when `phrase` is
NULL but `setting` is invalid is not practical to avoid without
breaking an existing guarantee, viz. that a "failure token" never
compares equal to the `setting` string, no matter what; I think that
guarantee is significantly more important than corner-case behavior
like this.

Note also that we have __nonnull annotations on all the pointer
arguments to all these functions, which means the C compiler
unconditionally treats passing NULL as any of those arguments as
provoking undefined behavior; I actually had to make crypt-port.h
suppress those annotations when building the library, in order to
get the tests to pass.  Possibly they should be removed.


crypt-md5.c
```diff
@@ -79,7 +80,7 @@ crypt_md5_rn (const char *phrase, const char *setting,
   const char *salt = setting;
 
   size_t salt_len;
-  size_t phrase_len;
+  size_t phrase_len = phr_size;
```
besser82: As `phrase_len` is used as a constant during the code, it should be replaced by directly using the passed parameter only.  The same goes for `salt_len` in the line above.

crypt-bcrypt.c
```diff
@@ -870,20 +870,21 @@ BF_crypt (const char *key, const char *setting, unsigned char *output,
...
-                 uint8_t *output, size_t o_size,
-                 void *scratch, size_t s_size)
+crypt_bcrypt_rn (const char *phrase, size_t ARG_UNUSED (phr_size),
+                 const char *setting, size_t ARG_UNUSED (set_size),
+                 uint8_t *output, size_t out_size,
+                 void *scratch, size_t scr_size)
 {
   /* This shouldn't ever happen, but...  */
-  if (o_size < BF_HASH_LENGTH || s_size < sizeof (struct BF_buffer))
+  if (out_size < BF_HASH_LENGTH || scr_size < sizeof (struct BF_buffer))
```
besser82: `set_size` must match `BF_SETTING_LENGTH` as well…

crypt-sha256.c
```diff
@@ -107,7 +109,7 @@ crypt_sha256_rn (const char *phrase, const char *setting,
   const char *salt = setting;
 
   size_t salt_len;
-  size_t phrase_len;
+  size_t phrase_len = phr_size;
```
besser82: Same as for `crypt-md5`…

crypt-sha512.c
```diff
@@ -107,7 +109,7 @@ crypt_sha512_rn (const char *phrase, const char *setting,
   const char *salt = setting;
 
   size_t salt_len;
-  size_t phrase_len;
+  size_t phrase_len = phr_size;
```
besser82: Same as for `crypt-md5`…

crypt-md5.c
zackw: I will replace phrase_len with phr_size (here and elsewhere).  salt_len, however, is not the same as set_size; if the setting string is a previously generated hash, it will be shorter.

crypt-bcrypt.c
zackw: Same deal here, BF_SETTING_LENGTH will be shorter than set_size if setting is a previously generated hash.

besser82: Allrighty, my bad.

Issue #18, opened 2018-07-24
Remove swapcontext use


fweimer:
I have doubts that this actually succeeds in wiping processor state. The downside is that it confusing valgrind a lot, and there are also compatibility issues with Intel CET.

zackw:
i'm working on this, for those very reasons.  I plan to replace it with a generalization of bcrypt's self-test.  However, I am not likely to get to it in the next few months.

PR #21, opened 2018-07-25
Remove use of makecontext/swapcontext

fweimer:
Fixes issue #18.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -72,7 +72,7 @@ AM_CONDITIONAL([HAVE_CPP_dD], [test $ac_cv_prog_cc_dD = yes])
 
 # Checks for header files.
 AC_CHECK_HEADERS_ONCE(
-  [fcntl.h stdbool.h sys/cdefs.h sys/random.h sys/syscall.h ucontext.h])
+  [fcntl.h stdbool.h sys/cdefs.h sys/random.h sys/syscall.h])
 
 # stdint.h and inttypes.h were checked for by AC_INCLUDES_DEFAULT,
 # invoked implicitly by the above.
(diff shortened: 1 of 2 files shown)


fweimer:
@besser82, what is needed to move this forward? Thanks.

(The alleged test coverage reduction is entirely spurious because we removed code which had full coverage.)

zackw:
Yeah, I dunno about codecov's percentage-change metric for patches.

I'm OK with removing the use of swapcontext, but I'm nervous about doing so without replacing it with anything.  In #18 I mentioned the possibility of implementing bcrypt-like selftests for all hashes as a substitute.  Is that a thing you would be willing to tackle, Florian?  I am not going to have time for it for a couple months.

fweimer:
@zackw I'm not sure that these selftests actually clobber all data if the hash is parameterized (because different code could run).

The guidance I have received is that password hashing is no longer supposed to cross a trust boundary, so this obfuscation shouldn't be required by extension. The current libxcrypt APIs are also not a good approach to avoid information leakage (because ideally, application code would *never* copy around passwords in cleartext, but that would need special read functions, and gets really complicated for very little benefit very quickly).

besser82:
A configure option would be nice.  So context swapping can be optionally disabled until we have a different solution…

zackw:
I thought about this some more and I've decided to withdraw my objections.  The security benefits of swapcontext are questionable, there are concrete problems caused by using it, and without it we're no worse off than any other existing implementation of crypt().  We can think about self-tests or other means of data sanitization later.  I will deal with the conflicts and merge the PR.

Issue #4, opened 2018-05-19
Feature request: configurable policy for hash functions and cost parameters


zackw:
On Linux, glibc's libcrypt forbids the use of MD5 and DES password hashes if `/proc/sys/crypto/fips_enabled` exists and contains a positive integer.  As I understand it, this was a kludge for a specific security certification that is no longer considered valuable, but it suggests a more general and useful feature: a configuration file (perhaps `/etc/security/pwhash.conf`) allowing the system administrator to forbid the use of some hash functions, and also to configure the options for hashes that have them.

A possible file format:

```
# As usual, comments begin with `#` and extend to the end of the line.
# Blank lines are ignored.
# New passwords are to be hashed in bcrypt-2b format with the default cost parameter
bcrypt-2b: yes

# SHA-2-512 is also acceptable.  Use 4096 rounds.
sha512: yes rounds=4096

# SHA-256 and older variants of bcrypt are acceptable but should be updated to a
# modern hash the next time the user logs in.
bcrypt-2a: old
bcrypt-2x: old
bcrypt-2y: old
sha256: old

# MD5 and DES hashes are no longer acceptable even for old accounts.
md5: no
des: no
bsdi: no
bigcrypt: no
```

`crypt` and `crypt_gensalt` would both fail with `EPERM` when offered a setting for a "no" hash; `crypt_gensalt` but not `crypt` would fail with `EPERM` when offered a setting for an "old" hash.  (Actually auto-updating the password database is not our job, but perhaps we could coordinate with the PAM people to make it happen.)

solardiz:
Not commenting on the feature nor on file format, but this reminded me: I suggest that we adopt the naming of these hashes that's currently used in password hash cracking. This means: descrypt, bsdicrypt, md5crypt, sha256crypt, sha512crypt. This would help avoid the usual confusion between DES (block cipher) and descrypt, and between MD5 (fast, non-salted non-password ex-cryptographic hash) and md5crypt, etc.

PR #26, opened 2018-08-21
[RFC] crypt.conf documentation

zackw:
Addressing issue #4, I've written up a bunch of documentation for a hypothetical new feature in which the set of hashing methods that may be used, and the default cost parameter for each, are configurable at runtime via a file `/etc/security/crypt.conf`.   I have not actually implemented the new feature and I don't know when I will have time to do it, but I thought it might be a good idea to invite feedback on the documentation first.  It will probably be easier to read if you check out the branch and then use `man -l` to read the formatted pages.  Start with `crypt.conf.5`.

This is not intended to be merged until the feature is actually implemented.

### DIFF
--- a/crypt-tune-costs.c
+++ b/crypt-tune-costs.c
@@ -1,0 +1,740 @@
+/* Benchmark hashing methods and set cost parameters appropriately.
+ *
+ * Written by Zack Weinberg <zackw at panix.com> in 2018.
+ * To the extent possible under law, the named authors have waived all
+ * copyright and related or neighboring rights to this work.
+ *
+ * See https://creativecommons.org/publicdomain/zero/1.0/ for further
+ * details.
+ */
+
+#include "crypt-port.h"
+
+#include <errno.h>
+#include <limits.h>
+#include <math.h>
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+
+enum hash_usage
+{
+  HU_PREFER, HU_ENABLED, HU_LEGACY, HU_DISABLED
+};
+
+static const char *const usage_keyword[] = {
+  "preferred", "enabled", "legacy", "disabled"
+};
+
+enum hash_cost_type
+{
+  HCT_EXPON, HCT_LINEAR, HCT_FIXED
+};
+
+struct hash_method
+{
+  const char *name;
+  const char *prefix;
+  enum hash_cost_type hct;
+  unsigned int minrounds;
+  unsigned int maxrounds;
+  enum hash_usage usage;
+  unsigned int nrounds;
+  double elapsed;
+};
+
+/* The 'usage', 'nrounds', and 'elapsed' fields of this table are
+   written to during execution.  */
+static struct hash_method hash_methods[] = {
+  { "yescrypt",      "$y$",   HCT_EXPON,  1,    11,         HU_PREFER,   0, 0 },
+  { "gost-yescrypt", "$gy$",  HCT_EXPON,  1,    11,         HU_ENABLED,  0, 0 },
+  { "scrypt",        "$7$",   HCT_EXPON,  6,    11,         HU_ENABLED,  0, 0 },
+  { "bcrypt",        "$2b$",  HCT_EXPON,  4,    31,         HU_ENABLED,  0, 0 },
+  { "bcrypt_a",      "$2a$",  HCT_EXPON,  4,    31,         HU_LEGACY,   0, 0 },
+  { "bcrypt_x",      "$2x$",  HCT_EXPON,  4,    31,         HU_LEGACY,   0, 0 },
+  { "bcrypt_y",      "$2y$",  HCT_EXPON,  4,    31,         HU_LEGACY,   0, 0 },
+  { "sha512crypt",   "$6$",   HCT_LINEAR, 1000, 999999999,  HU_ENABLED,  0, 0 },
+  { "sha256crypt",   "$5$",   HCT_LINEAR, 1000, 999999999,  HU_ENABLED,  0, 0 },
+  { "sha1crypt",     "$sha1", HCT_LINEAR, 4,    4294967295, HU_LEGACY,   0, 0 },
+  { "sunmd5",        "$md5",  HCT_LINEAR, 4096, 4294963199, HU_LEGACY,   0, 0 },
+  { "md5crypt",      "$1$",   HCT_FIXED,  0,    0,          HU_LEGACY,   0, 0 },
+  { "bsdicrypt",     "_",     HCT_LINEAR, 1,    16777215,   HU_DISABLED, 0, 0 },
+  { "bigcrypt",      "",      HCT_FIXED,  0,    0,          HU_DISABLED, 0, 0 },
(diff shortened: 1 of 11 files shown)


besser82 (commented):
LGTM, with a small remark about only using `/etc` for the preset file:

Files placed in some dir under `/etc` are meant to be edited by admins and shouldn't be simply clobbered by a distribution preset, if defaults change and all.

Maybe there should be a distribution specific config located in `/usr/lib` as it is done by e.g. `rpm` or `systemd` amoung some others.  The setting from this file should be overridden by the settings from the file in `/etc`.

That way distributions can change defaults in an easy way by leaving a way to admins that decide to diverge from distro presets.

rfc1036:
Debian maintainer here.

I think that this kind of scheme (the file in /usr is ignored if there is one in /etc)  would be worse in this case, because then administrators who create the /etc file to modify a specific setting would not be notified of possibly important changes in the /usr file in future releases.

Debian and Ubuntu have well established methods to preserve and manage changes to files in /etc/, and RPM systems at least warn at package installation time if a configuration file has changed.

besser82:
@rfc1036 I've never said not to issue a warning on package installation / upgrade.  I'm just saying there should be an option for a sane, distro specific preset.

In the case of rpm the file in `/etc` would be marked as `%config(noreplace)` and would issue a warning and create a `.rpmnew`-file in the same loaction, if the file in the new package differs.

This would inform the admin properly, but leave the old config in effect as long as the file in `/etc` isn't updated.  This ensures there is no fallout on running systems.

zackw:
In this case, I was imagining that there would be no file installed by our
“make install”; the library would just use its compiled-in defaults.
However, package postinstall scripts could run crypt-tune-costs
automatically. Since that would produce a file customized for the specific
machine it was run on, /etc seems like the proper home for it.

thkukuk:
/etc/security is a bad choice, as this is used for PAM configuration files. Putting other stuff there is very confusing.

And putting the configuration file in /usr: nearly every distribution, if not even all, who support atomic updates, put the distribution specific configuration file in /usr: /usr/etc, /usr/share/default, /usr/lib/basesystem, and a lot of more. The reason is, that /etc is not accessible on such systems during upgrade. There are three ways how they use this: put a symlink in /etc pointing to /usr, which the admin can replace with a copy with his changes. Do a three-way-merge at next reboot. Not really reliable. Or, preferred, use something like systemd is doing: /usr contains the distribution config file, and /etc only contains the changes from the admin. So distributions can put easily new options in /usr without the risk that the configuration breaks after update.

zackw:
@thkukuk Where in `/etc` should a machine-specific crypt.conf go, then?  Just loose at top level?

besser82:
I'd say it should be configurable below `/etc`.  Depending on the distribution there are different dirs for single, additional config files:

* Debian-ish:  `/etc/default/crypt.conf`
* redhat-ish:  `/etc/sysconfig/crypt.conf`

thkukuk:
On Sat, Sep 08, Björn Esser  wrote:

> I'd say it should be configurable below /etc. Depending on the distribution
> there are different dirs for single, additional config files:
> 
>   • Debian-ish: /etc/default/crypt.conf
>   • redhat-ish: /etc/sysconfig/crypt.conf

Both, /etc/default and /etc/sysconfig, contain by definition environment
files, which are sourced bei init scripts or systemd services and contain
options for the daemons. This don't sound like the correct directory.

On SUSE (which is also conform with FHS) we put such single configuration
files still directly in /etc, subdirectories should only be used or 
created, if a services has multiple configuration files.
What's wrong with that? That's where admin normally looks for such files.

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CaaSP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

rfc1036:
As the Debian maintainer I agree: I would just use `/etc/crypt.conf` and this way we would use the same name among different distributions.

`/etc/defaults/` is mostly hacks to work around daemons which lack configuration files, and I feel that nowadays it is vaguely deprecated in Debian.

solardiz:
This PR introduces crypt.conf names for the various hash types, and we'll be stuck supporting those names. Let's be more careful in how we choose them. Instead of what's currently proposed:

```
sha512
sha256
sha1
md5_sun
md5_bsd
des_bsd
des_big
des
nthash
```

I strongly recommend the following:

```
sha512crypt
sha256crypt
sha1crypt
sunmd5
md5crypt
bsdicrypt
bigcrypt
descrypt
nt
```

This is the naming currently used in JtR. Well, for "nthash" vs. "nt" I don't feel strongly, but for the rest it's important that we do not invent new names nor proliferate the confusion between e.g. sha512crypt and raw SHA-512 nor between e.g. descrypt and DES. I understand the desire for consistent naming based on the underlying cryptographic primitive, but we can't do that consistently anyway (e.g., bcrypt is better known as bcrypt, and not as Blowfish), it's confusing to many users who are not familiar with how very different these high-level algorithms are from their underlying cryptographic primitives, and in the end the high-level algorithms around whatever primitive typically matter more than the choice of primitive. I had originally made the same mistake in JtR in calling things "des" and "md5" and "bf", and that was confusing people. These have since (a few years ago) been renamed to "descrypt" and "md5crypt" and "bcrypt", etc.

Perhaps we should also edit the sub-headings in the man page to use this or similar naming, but we can revise those any time without worrying about backwards compatibility. So only fixing the crypt.conf names as above is a must before merging this PR.

zackw:
@solardiz Thanks for the comprehensive list of JtR's names.  I will make this change next time I cycle around to working on this project.

crypt-tune-costs.c
```diff
@@ -0,0 +1,740 @@
...
+              usage_keyword[hash_methods[i].usage]);
+
+  if (fflush (stdout) || fclose (stdout))
+    {
+      fprintf (stderr, "%s: stdout: %s\n", program_name, strerror (errno));
+      exit (1);
+    }
+}
+
+static NORETURN
```
besser82: You should use the [`format` attribute](http://clang.llvm.org/docs/AttributeReference.html#format) here, so Clang doesn't complain about `-Wformat-nonliteral`.  Using `__attribute__((__format__ (__printf__, 1, 0)))` should be sane according to the [GCC online docs](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes).

besser82: Resolved in https://github.com/besser82/libxcrypt/pull/26/commits/4953f92c0db7eb45d83dd293635bf12f70799839.

besser82:
@zackw You may squash or edit any of my commits as needed.

besser82:
@zackw FYI, I've rebased this onto recent `develop`.

besser82:
@zackw Rebased against recent `develop`.

zackw:
I don't have time to work on this project myself in the near future, but, looking at #117, #130, etc. I think it should be a high priority for anyone interested in hacking on libxcrypt.

vortexilation:
@zackw 

 5 years later, Is there any continuation about this PR?.

PR #31, opened 2018-09-07
Full backend for scrypt with tests

besser82:
As mentioned in #27 here comes the backend for scrypt.

### DIFF
--- a/crypt-port.h
+++ b/crypt-port.h
@@ -108,8 +108,8 @@ typedef union
   explicit_memset (s, 0x00, len)
 #else
 /* The best hope we have in this case.  */
-static inline
-void _xcrypt_secure_memset (void *s, size_t len)
+static inline void
+_xcrypt_secure_memset (void *s, size_t len)
 {
   volatile unsigned char *c = s;
   while (len--)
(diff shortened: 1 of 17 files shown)


crypt-port.h
```diff
@@ -119,10 +119,25 @@ void _xcrypt_secure_memset (void *s, size_t len)
...
+  assert (src != NULL);
+  const size_t s_size = strlen (src);
+  assert (d_size >= s_size + 1);
+  memcpy (dst, src, s_size);
+  XCRYPT_SECURE_MEMSET (dst + s_size, d_size - s_size);
+  return s_size;
+}
+#define XCRYPT_STRCPY_OR_ABORT(dst, d_size, src) \
+  _xcrypt_strcpy_or_abort ((char *) dst, (const size_t) d_size, \
+                           (const char *) src)
```
zackw: Why are the casts in the definition of `XCRYPT_STRCPY_OR_ABORT` necessary?  If it's because some callers pass `char *`, some pass `unsigned char *`, and some pass `void *`, instead I suggest making the inline function take `void *` and `const void *` arguments instead, and then casting in the body of the function.  That way the compiler will at least still catch const-correctness errors.

zackw: I'm going to go ahead and merge this and then tinker with `XCRYPT_STRCPY_OR_ABORT` myself.

PR #34, opened 2018-09-10
Add version macros

besser82:
Macros for conveniant version checks.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -12,6 +12,7 @@ AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_FILES([
   Makefile
   libxcrypt.pc
+  crypt-base.h
 ])
 AM_INIT_AUTOMAKE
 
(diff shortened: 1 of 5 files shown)


configure.ac
```diff
@@ -333,4 +334,17 @@ AM_CONDITIONAL([ENABLE_SHARED], [test $enable_shared = yes])
 # The Makefile needs to know if we are cross-compiling.
 AM_CONDITIONAL([CROSS_COMPILING], [test $cross_compiling = yes])
 
+XCRYPT_VERSION_MAJOR=$(
+    echo $PACKAGE_VERSION | $AWK '{split($0,a,"."); print a@<:@1@:>@}'
```
zackw: This can be done more simply: `| $AWK -F. '{ print $1 }'` or `| cut -d. -f1`.  And then you don't need to use `@<:@`.

besser82:
Well, maybe it might be a good idea to introduce an `uint32_t xcrypt_supports (const char* setting)` function with a corresponding `enum {crypt = 1 << 0, gensalt = 1 << 1} method_supported`?

zackw:
Yes, I think a runtime "do we support this hash?" query function would be a good idea.  There's some overlap with the [`crypt_checksalt`](https://github.com/besser82/libxcrypt/blob/2c908d006a87cbada3b6fef5c82739efdb92e19b/crypt_checksalt.3) function I proposed in conjunction with `crypt.conf` -- I think one function could be made to do both jobs.

besser82:
Then we should implement `crypt_checksalt()` step-by-step, starting with `CRYPT_CHECKSALT_IMPLEMENTS_METHOD_AVAILABLE`.  I'll start working on that tomorrow.  I'll drop the version comparison during that.

besser82:
[Travis](https://travis-ci.org/besser82/libxcrypt/builds/426901988) is fine, comments have been resolved. Merging on cli.

PR #35, opened 2018-09-10
Remove sha256crypt ($5$) and sha512crypt ($6$) from strong set.

zackw:
These are the strongest hashes supported by glibc’s libcrypt, but
they have no particular protection against hardware parallel
attacks (e.g. using GPUs to iterate SHA-2) and should probably
only be used for new hashes in mixed environments where yescrypt
and/or bcrypt aren’t universally supported.

It’s difficult to find information about how difficult these hashes
specifically are to brute-force; the big online services whose user
databases got dumped seem to have skipped straight from SHA-1 or worse
straight to bcrypt.  I think what I said in NEWS is accurate as far as
it goes, but it might be nice to add more concrete numbers and advice
to crypt.5.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,6 +11,10 @@ Version 4.2.0
 * --enable-hashes now supports additional groups of hashing methods:
   'altlinux', 'owl', and 'suse', which select the methods historically
   supported on those operating systems.
+* --enable-hashes=strong now excludes the SHA-2-based hashes
+  ($5$ and $6$).  These are still cryptographically sound, but not
+  hardened against massively parallel brute-force attacks (e.g. using
+  GPUs to compute SHA-2) as the newer bcrypt, scrypt, and yescrypt are.
 
 Version 4.1.2
 * Add optional 'check-valgrind' target to the Makefile.
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ supported hashes.  Disabling the traditional ‘des’ hash algorithm
 implies `--disable-obsolete-api`.  Security-conscious environments
 without backward compatibility constraints are encouraged to use
 `--enable-hashes=strong`, which enables only the hash functions that
-are strong enough to be safe for newly hashed passwords.
+are definitely strong enough to be safe for newly hashed passwords.
 
 The original implementation of the SUNMD5 hashing algorithm has a bug,
 which is mimicked by libxcrypt to be fully compatible with hashes
--- a/crypt.5
+++ b/crypt.5
@@ -166,6 +166,7 @@ originally for the Tarsnap online backup service.
 The algorithm was specifically designed to make it costly to perform
 large-scale custom hardware attacks by requiring large amounts of memory.
 In 2016, the scrypt algorithm was published by IETF as RFC 7914.
+Not as strong as yescrypt, but still acceptable for new hashes.
 .hash "$7$" "\e$7\e$[./A-Za-z0-9]{11,97}\e$[./A-Za-z0-9]{43}" unlimited 8 256 256 "up to 512" "6 to 11 (logarithmic)"
 .PP
 .ti -4
@@ -176,6 +177,7 @@ modified to have an extra-expensive key schedule.
 Originally developed by Niels Provos and David Mazieres for OpenBSD
 and also supported on recent versions of FreeBSD and NetBSD,
 on Solaris 10 and newer, and on several GNU/*/Linux distributions.
+Not as strong as yescrypt, but still acceptable for new hashes.
 .hash "$2b$" "\e$2[abxy]\e$[0-9]{2}\e$[./A-Za-z0-9]{53}" 72 8 184 184 128 "4 to 31 (logarithmic)"
 .PP
 The alternative prefix "$2y$" is equivalent to "$2b$".
@@ -190,7 +192,7 @@ which incorrectly processed characters with the 8th bit set.
 A hash based on SHA-2 with 512-bit output,
 originally developed by Ulrich Drepper for GNU libc.
 Supported on Linux but not common elsewhere.
-Acceptable for new hashes.
+Not recommended for new hashes, but not catastrophically weak.
 The default CPU time cost parameter is 5000,
 which is too low for modern hardware.
 .br
@@ -202,7 +204,7 @@ which is too low for modern hardware.
 A hash based on SHA-2 with 256-bit output,
 originally developed by Ulrich Drepper for GNU libc.
 Supported on Linux but not common elsewhere.
-Acceptable for new hashes.
+Not recommended for new hashes, but not catastrophically weak.
 The default CPU time cost parameter is 5000,
 which is too low for modern hardware.
 .br
--- a/hashes.lst
+++ b/hashes.lst
@@ -34,8 +34,8 @@ bcrypt   _a         $2a$      16       STRONG,ALTLINUX,FREEBSD,NETBSD,OPENBSD,OW
 bcrypt   _b         $2b$      16       STRONG,ALTLINUX,FREEBSD,NETBSD,OPENBSD,OWL,SOLARIS,SUSE
 bcrypt   _x         $2x$      16       STRONG,ALTLINUX,FREEBSD,NETBSD,OPENBSD,OWL,SOLARIS,SUSE
 bcrypt   _y         $2y$      16       STRONG,ALTLINUX,FREEBSD,NETBSD,OPENBSD,OWL,SOLARIS,SUSE
-sha512   :          $6$       15       STRONG,GLIBC,FREEBSD,SOLARIS
-sha256   :          $5$       15       STRONG,GLIBC,FREEBSD,SOLARIS
+sha512   :          $6$       15       GLIBC,FREEBSD,SOLARIS
+sha256   :          $5$       15       GLIBC,FREEBSD,SOLARIS
 md5      :          $1$       9        GLIBC,FREEBSD,NETBSD,OPENBSD,SOLARIS
 sunmd5   :          $md5      8        SOLARIS
 sha1     :          $sha1     20       NETBSD


rfc1036:
https://gist.github.com/epixoip/a83d38f412b4737e99bbef804a270c40 is a good benchmark, even if a bit old, and it shows that in practice sha512crypt can be cracked 10x faster than bcrypt.
So in the general case sha-512 should be as good as bcrypt as long as 10x more rounds are used: the glibc default is 5000.

solardiz:
All of these hashes are tunable, so ideally any benchmarks we would consider here should be for defensive performance vs. cracking performance, or for cracking performance at similar and currently relevant defensive throughput (unless we assume that anyone using SHA-crypt would use the default of rounds=5000 or else migrate to another hash type? maybe). The benchmarks cited above are cracking-only at tunable settings that are historically used for benchmarking, but that have somewhat different defensive throughput (SHA-crypt's default of rounds=5000, and bcrypt at cost 5 unless stated otherwise), so direct/exact comparisons of these performance numbers between different hash types are at best not ideal.

Besides the performance aspects, a drawback of SHA-crypt (and especially sha256crypt) is its running time being highly dependent on password length. For unusually high lengths, it's O(length^2). For low lengths, in sha256crypt there are very noticeable (tens of percent) differences in running time for realistic and attack-relevant password length ranges - e.g., for 7 vs. 8 or 11 vs. 12 characters, with the exact thresholds varying by salt length. (For sha512crypt, similar slowdowns occur at much higher lengths - tens of characters - so not practically relevant as an infoleak.) This also limits the cost setting a sysadmin or distro vendor could realistically use - would have to consider the worst case (longest) allowed passwords, and as a result have unnecessarily high throughput (including for cracking) for more realistic lengths.

There's also a relatively tiny timing leak (in both sha256crypt and sha512crypt) of 8 bits of an intermediate hash value (independent of password length), so precise timing measurements would allow an attacker to rule out 255 out of 256 possible passwords (or better if also considering lengths), but I think this attack will remain impractical and thus only theoretical (although it can be demo'ed).

bcrypt, scrypt, and yescrypt are not ideal in that they're cache-timing unsafe. That's a tough security trade-off (avoiding this property makes the password hash weaker in other ways). But they avoid "worse than cache" timing leaks, whereas SHA-crypt literally has conditional branching based on sensitive data (not limited to password length - also on that 8-bit intermediate hash).

solardiz:
The proposed change to NEWS (indirectly) calls bcrypt "newer", whereas it actually pre-dates SHA-crypt by a decade. For the same reason, bcrypt wasn't deliberately hardened against GPUs - those were quite different and were not used for password cracking at the time bcrypt was introduced. It just happened to be GPU-unfriendly (at first very unfriendly, but newer GPUs gradually become better at it).

I am just sharing what I know and notice. I do not yet vote for/against a change like what's proposed here.

solardiz:
For our current results at SHA-crypt on FPGA, see:

https://www.openwall.com/lists/john-users/2018/07/23/1 (sha512crypt)
https://www.openwall.com/lists/john-users/2018/08/27/11 (sha256crypt)

These are much worse than the corresponding latest GPU speeds per chip, but are on par with latest GPUs per Watt. However, they're on FPGAs from many years ago (45nm, circa 2011).

zackw:
I had been under the impression that sha{256,512}crypt were known to be significantly weaker than bcrypt at any cost parameter level -- it seems I was wrong.  Let's close this PR for now and make a note that we need to collect solid data to base any decision on.

The branch for #26 now contains a sketch implementation of a program called `crypt-tune-costs` that's meant to select appropriate cost parameters for the machine it's run on, but all it does is measure what cost parameter makes each hash take some amount of wall-clock time (defaulting to 250ms).  I'm not even sure this is the right thing for the stated purpose of per-installation tuning, but perhaps it can inform this discussion a little?  On my laptop, these are the numbers I get:

```
hash         cost     elapsed
----         ----     -------
yescrypt     10       316.05ms
scrypt       9        355.83ms
bcrypt       13       404.67ms
sha512crypt  293877   250.23ms
sha256crypt  266405   253.79ms
sha1crypt    270947   252.51ms
md5_sun      175569   251.48ms
des_bsd      1966043  251.21ms
```

(if you run the program yourself, pass `--enabled=all`, the output will be in a slightly different format, and don't worry about what "enabled" vs "legacy" means)

solardiz:
> https://gist.github.com/epixoip/a83d38f412b4737e99bbef804a270c40 is a good benchmark, even if a bit old

We don't have to use that one, Jeremi has since posted a newer one for 8x1080Ti and newer hashcat:

https://gist.github.com/epixoip/ace60d09981be09544fdd35005051505

He also posted bcrypt benchmarks on 4 different recent GPUs:

https://gist.github.com/epixoip/9d9b943fd580ff6bfa80e48a0e77520d

solardiz:
Tuning for same defensive running time on one logical CPU on an idle system isn't ideal. A more important figure (limiting the maximum cost setting that can be used on a server) is throughput (hashes computed per second) under full CPU utilization (all logical CPUs) by concurrent hash computations. But the differences between these two metrics are usually within 2x (of course, first taking into consideration that they're inversely proportional), so I'll just use @zackw's results below. Also, I don't know what password length and salt these benchmarks are for - and this matters a lot for sha256crypt - but again usually within a factor of 2. hashcat's are for password length 7, salt length 8.

Finally, I think 250ms on a modern CPU is excessive as a distro's default (more useful is the range up to 100ms), and with our current single-parameter input to yescrypt would consume more memory (great for security) than a distro would dare to use for this task by default (considering it's per each concurrent authentication attempt, so can be multiple GB total). Thankfully, for the 4 hash types I reference below this isn't a problem as it relates to scaling the existing results.

Based on Jeremi's 8x1080Ti benchmarks I referenced above, we can expect these speeds at the cost settings from @zackw's comment above:

| hash | 8x1080Ti cracking h/s per 250ms on @zackw's laptop |
| --- | --- |
| bcrypt | 454 |
| sha512crypt | 31.4k |
| sha256crypt | 89.6k |
| bsdicrypt | 7.1k |

No directly comparable data for other hash types @zackw listed.

Scaling formulas I used:

```
184.8*1000*(2^5*1042+585)/(2^13*1042+585)*250/404.67 = 453.76
1849.1*1000*5000/293877 = 31460
4774.6*1000*5000/266405 = 89612
19333.5*1000*725/1966043 = 7129
```

Per these results, SHA-crypt and even bsdicrypt certainly do appear a lot cheaper than bcrypt to attack on GPUs. However, per the speed for bsdicrypt I guess hashcat's implementation of it is not bitsliced (sorry, too lazy to check the code). Using the number for descrypt (which I know is bitsliced) instead we'd get:

```
11414.2*10^6*25/1966043 = 145142
```

as the potential speed for a properly optimized implementation. This needs to be scaled down somewhat since bsdicrypt will require extra registers for the pointers to extra 12 bits of salt (24-bit vs. 12-bit salt), and register pressure is often the limiting factor for performance of bitslice implementations. So maybe 100k, which is on par with sha256crypt.

So bcrypt does stand out.

On FPGAs, the results are very different (the opposite as it relates to bcrypt vs. sha512crypt), but GPUs are still far more accessible (this might change somewhat as/if we proceed to implement bcrypt on AWS F1).

solardiz:
| hash | one ZTEX 1.15y board (4 FPGAs) cracking h/s per 250ms on @zackw's laptop |
| --- | --- |
| bcrypt | 682 |
| sha512crypt | 936 |
| sha256crypt | 2.5k |
| bsdicrypt | 10k |

Scaling formulas I used:

```
106000*(2^5*1042+585)/(2^13*1042+585)*404.67/250 = 682
55000*5000/293877 = 936
133000*5000/266405 = 2496
800*10^6*25/1966043 = 10173
```

We don't have an implementation of bsdicrypt on FPGA, so I used scaling from descrypt - it's similar enough.

The bcrypt speed will probably be a little better than what I computed here since the overhead is higher at lower cost settings. I'll try to run a direct cost 13 test, or scale from a higher cost.

Nevertheless, it looks like bcrypt wins even at this test.

zackw:
Here's what I get for target times of 100ms, on the same computer:

```
yescrypt     9        157.85ms
scrypt       8        176.88ms
bcrypt       11       100.22ms
sha512       118781   100.40ms
sha256       105841   100.03ms
sha1         107239   100.09ms
md5_sun      67433    100.08ms
des_bsd      789397   100.21ms
```

and 50ms:

```
yescrypt     8         78.11ms
scrypt       7         89.30ms
bcrypt       11       100.11ms
sha512       59723     50.31ms
sha256       53743     50.96ms
sha1         54495     50.91ms
md5_sun      10001     50.17ms
des_bsd      392947    50.11ms
```

For exponential cost parameters, it chooses the smallest number producing a time above the target.

I will think about revising the program to estimate throughput instead, but I'm not sure when I will have time to implement it.

solardiz:
Thanks @zackw. These results scale for the 4 hash types I included in my comparison as expected, so I don't need to redo my calculations.

solardiz:
> I will think about revising the program to estimate throughput instead, but I'm not sure when I will have time to implement it.

While throughput is more relevant for defense vs. attack comparisons, it makes sense to keep tuning for elapsed time as a feature as well. Ideally, it should be possible to tune for either throughput or elapsed time (command-line options?), and have the other metric output as well. Or maybe it'll be more metrics to output: also elapsed time when running 1 thread vs. when running max threads. Could also output min/avg/max like I do in yescrypt's `userom.c` program (which I also hope to find time to rewrite one day):

```
Benchmarking 1 thread ...
100 c/s real, 100 c/s virtual (255 hashes in 2.54 seconds)
Benchmarking 56 threads ...
2314 c/s real, 41 c/s virtual (14280 hashes in 6.17 seconds)
min 13.440 ms, avg 24.049 ms, max 25.467 ms
```

(In this example, "min" is the lowest seen when running 56 threads, and it just happens to be similar to what would be typical for 1 thread. We should probably also have explicit min/avg/max output for 1 thread.)

zackw:
It looks like the permissively-licensed SHA-256 and -512 may be slightly faster (needing more iterations for the same target time) and MD5 might be slightly slower.  But I don't know whether any of the changes are statistically significant -- that would require me to do more complicated testing than I have time for today (or in the near future).

```
hash           cost     elapsed
----           ----     -------
yescrypt         10	301.61ms
scrypt            9	340.82ms
bcrypt           13	385.78ms
sha512crypt  327849	254.29ms
sha256crypt  303227	252.29ms
sha1crypt    263921	251.55ms
md5_sun      166999	252.10ms
des_bsd     1819753	254.74ms
```

solardiz:
Thanks, @zackw. Those changes don't appear to be statistically significant - there's also a similar change for `des_bsd`, which isn't supposed to be affected by the code changes. At least we know the differences are within 10% or so.

Also, this reminds me: we're probably wasting time on context zeroization on every iteration of sha*crypt and SunMD5, because it's done in the primitives' `*_Final()`. We might want to look into that and fix it under a separate issue... or not, if we keep those for backwards compatibility only.

solardiz:
6fc8102f7727f8e4115a87070d2e4a36b55ea1bc should indeed hurt performance at MD5 and MD4 a lot. Any chance we can revert it? @besser82

zackw:
That's doing aliasing-unsafe pointer tricks. I want all such code
exterminated from the repository *even if* that hurts performance.

solardiz:
OK, it's your call. I understand that's an aliasing violation, as I acknowledge in the comment. It's prevented from turning into a problem now by the translation unit boundary, which might not work that way forever. To avoid the violation yet not incur the performance hit, we'd have to modify the context API to accept a `union` and modify the caller accordingly. I guess this is in fact not worth the bother for the legacy uses like SunMD5.

zackw:
Yeah, I'd think harder about preserving this kind of optimization in
the code used by bcrypt or yescrypt, but not anything depending on
MD[45] or even SHA1.

PR #36, opened 2018-10-18
Convert existing manpages to BSD mdoc format.

zackw:
I've been looking into the possibility of generating HTML documentation from our manpages, to be hosted on Github Pages or something similar.  I'm not sure when I will get around to finishing that project, but as a first step I would like to propose we convert the manpages from traditional tmac.an format to the newer tmac.doc format, and I'd like to merge that now, so that I don't have a long-lived branch whose first changeset involves a big messy reformatting operation that's very likely to conflict with future changes to these files.

mdoc is much more semantic than man, and is therefore a better starting point for generation of HTML pages.  Also, the “mandoc” program for generating HTML from mdoc source is actually maintained, unlike “man2html.”  GNU troff has shipped tmac.doc for many years, so I don't think this should pose a compatibility problem with any system where libxcrypt is likely to be used.

This conversion deliberately ignores some of the documented mdoc conventions: in particular, I think SEE ALSO belongs at the end, and it’s better to put the pages in the same section as the current page first in the SEE ALSO list.

This conversion also winds up rendering slightly better with GNU “nroff -mdoc” than with mandoc, because mandoc doesn’t implement macros inside tbl cells; this is a known missing feature that’s already on their to-do list.

### DIFF
--- a/crypt.3
+++ b/crypt.3
@@ -1,426 +1,468 @@
 .\" Written and revised by Solar Designer <solar at openwall.com> in 2000-2011.
 .\" Revised by Zack Weinberg <zackw at panix.com> in 2017.
+.\" Converted to mdoc format by Zack Weinberg in 2018.
 .\"
 .\" No copyright is claimed, and this man page is hereby placed in the public
 .\" domain.  In case this attempt to disclaim copyright and place the man page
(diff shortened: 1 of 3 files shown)


besser82:
LGTM.  Will merge on cli.

PR #38, opened 2018-10-24
yescrypt: Add more tests.

vt-alt:
Test that crypt() with hashed password in place of settings works the
same as settings and data after settings are ignored.

@ldv-alt @solardiz


besser82:
LGTM!  Thank you! =)

PR #44, opened 2018-10-29
Use macros to detect endianness if configure cannot tell.

### LINKED ISSUE #42: error: 'IS_BIGENDIAN' is not defined, evaluates to 0
After fixing the other issues I reported, libxcrypt fails to build with this error—if I build universal:

```
alg-sha1.c:84:5: error: 'IS_BIGENDIAN' is not defined, evaluates to 0 [-Werror,-Wundef]
#if IS_BIGENDIAN
    ^
1 error generated.
make[1]: *** [libcrypt_la-alg-sha1.lo] Error 1
```

The configure script is trying to `#define IS_BIGENDIAN 1` on big-endian systems and `#define IS_BIGENDIAN 0` on little-endian systems, but autoconf has a third value for its endian check when building universal:

```
checking whether byte ordering is bigendian... universal
```

A [universal binary](https://en.wikipedia.org/wiki/Universal_binary) is a macOS feature that allows a developer to distribute a single binary that runs on any Mac regardless of what processor it uses. Building universal is ideally accomplished by running `./configure` a single time, supplying multiple `-arch` flags in `CFLAGS`, `CXXFLAGS`, `LDFLAGS` etc., then running `make` and `make install` a single time. When invoked with multiple `-arch` flags, the compiler internally runs itself once per architecture and then glues the results into a single file using `lipo`.

This method requires you not to attempt to detect endianness or bitness at configure time. Autoconf's universal "support" was designed when "universal" meant building for 32-bit PowerPC and 32-bit Intel (e.g. `-arch i386 -arch ppc`)—the bitness was the same but the endianness differed. When autoconf detects that multiple `-arch` flags are used, it therefore answers that the endianness is "universal" to ensure you don't rely on this result. On modern systems, "universal" means building for 32-bit Intel and 64-bit Intel (e.g. `-arch i386 -arch x86_64`)—the endianness is the same but the bitness differs—but autoconf has not been updated to detect that the endianness is the same, so it still answers "universal".

A less-ideal way of building universal is to create two build directories, and in one, `./configure`, `make` and `make install` with one arch, do the same with the other arch in the other directory, and then manually `lipo` any binary files together, and I can do that if necessary, but it would be great if it weren't.

I recommend removing the `IS_BIGENDIAN` stuff entirely and instead use `#ifdef __BIG_ENDIAN__`; compilers on macOS have that predefined for you. And if you want to know if you're on 64-bit vs 32-bit, you can check `#ifdef __LP64__`. I don't know how well supported these defines are on compilers used on non-Mac platforms.

besser82:
Some systems provide a way to build a single binary that runs on any of their architectures.  Thus detecting the system's endianness statically during configure stage is not sufficient.

Fixes #42.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -47,11 +47,12 @@ LT_INIT
 
 # Check for system's endianness
 AC_C_BIGENDIAN(
-   AC_DEFINE([IS_BIGENDIAN], 1,
-    [Define to 1 if system's architecture is big-endian.]),
-   AC_DEFINE([IS_BIGENDIAN], 0,
-    [Define to 0 if system's architecture is little-endian.])
+  AC_DEFINE(XCRYPT_BE_ARCH, 1, [System is bigendian.]),
+  AC_DEFINE(XCRYPT_LE_ARCH, 1, [System is littleendian.]),
+  AC_MSG_WARN([Relying on macros defined in "crypt-port.h" to detect system's endianness.]),
+  AC_MSG_WARN([Relying on macros defined in "crypt-port.h" to detect system's endianness.])
 )
+
 # One of the test scripts needs to use -dD.
 AC_CACHE_CHECK([whether the preprocessor ($CPP) supports -dD],
   [ac_cv_prog_cc_dD],
--- a/crypt-port.h
+++ b/crypt-port.h
@@ -62,6 +62,83 @@
 #define MIN_SIZE(x) (x)
 #endif
 
+/* Macros for detecting endianness of the system at compile time.  */
+#if !(defined XCRYPT_BE_ARCH || defined XCRYPT_LE_ARCH) && \
+     (defined __ARMEB__ || defined __THUMBEB__ || defined __AARCH64EB__ || \
+      defined _MIPSEB || defined __MIPSEB || defined __MIPSEB__ || \
+     (defined __BYTE_ORDER__ && defined __ORDER_BIG_ENDIAN__ && \
+         __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__) || \
+     (defined __BYTE_ORDER__ && defined __BIG_ENDIAN__ && \
+         __BYTE_ORDER__ == __BIG_ENDIAN__) || \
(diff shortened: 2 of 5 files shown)


crypt-port.h
```diff
@@ -68,6 +68,81 @@
 #define MIN_SIZE(x) (x)
 #endif
 
+/* Macros for detecting endianness of the system at compile time.  */
```
zackw: Hm.  Where did you get this big list of macros?  How do we know it's exhaustive? Is there any way we could plausibly make it shorter?  Two possiblities come to mind: make use of `sys/endian.h` where available, and keep the autoconf check around and limit the cases in the if statement to the ones that are relevant when `IS_BIGENDIAN` is undefined.

besser82: I've re-added and modified the configure check in the rebased commit.  The macros are used only when configure sets endianness to `universal`.

BTW, most of the macros are comming from https://sourceforge.net/p/predef/wiki/Endianness/ and the variations of underscores were added by me, since some platforms define those macros in different variations.

besser82:
Now the macOS build should be fine except for the linker invocation, but that will be fixed in a different PR.

configure.ac
```diff
@@ -47,11 +47,12 @@ LT_INIT
...
 # Check for system's endianness
 AC_C_BIGENDIAN(
-   AC_DEFINE([IS_BIGENDIAN], 1,
-    [Define to 1 if system's architecture is big-endian.]),
-   AC_DEFINE([IS_BIGENDIAN], 0,
-    [Define to 0 if system's architecture is little-endian.])
+  AC_DEFINE(XCRYPT_BE_ARCH, 1, [System is bigendian.]),
+  AC_DEFINE(XCRYPT_LE_ARCH, 1, [System is littleendian.]),
+  AC_MSG_ERROR([Cannot compile libxcrypt on a system with unknown endianness!]),
+  AC_MSG_WARN([Relying on macros defined in "crypt-port.h" to detect system's endianness.])
```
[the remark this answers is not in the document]

besser82:
@zackw Will merge when TravisCI goes green.

PR #47, opened 2018-10-30
Add generic test if crypt is accepting gensalt and crypt outputs

vt-alt:
Some backends do not test if a hashed password is accepted properly as a
setting, this is a generic place to test it for all backends.

Only default rounds size is tested otherwise the test is taking too long.

### DIFF
--- a/test-gensalt.c
+++ b/test-gensalt.c
@@ -464,7 +464,37 @@ main (void)
                      tcase->prefix, tcase->rounds, ent, salt);
 
           XCRYPT_STRCPY_OR_ABORT (prev_output, CRYPT_GENSALT_OUTPUT_SIZE, salt);
-        }
+
+	  /* Test if crypt works with this salt. */
+	  if (!tcase->rounds)
+	    {
+#define PASSW "alexander"
+	      static struct crypt_data a, b;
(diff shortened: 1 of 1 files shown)


vt-alt:
> 1 failing check

I think this error should be ignored in regard to this patch, as it's a bug (or feature?) of bigcrypt, their `gensalt_bigcrypt_rn` output (just 2 bytes, as in `gensalt_descrypt_rn`) is not accepted by `crypt_bigcrypt_rn` (requiring 14 bytes). I don't know how it supposed to be used...

besser82:
@vt-alt Can you rebase against recent `develop` branch, please?  The failing test should be fixed then.

vt-alt:
If test password is "default" (<= 8 chars) then resulting hash for setting `Mp............` is `Mpe4CDQr4ziYM`. It is 13 characters and fails `if (set_size <= 13)`.

Is test password is set to be "alexander" (> 8 char), then for the same setting resulting hash is `Mpbn1WoPbF.LYLXs6Vco3KFQ`.

vt-alt:
Probably, I should just set test pass to "alexander" and rebase.

PR #49, opened 2018-11-01
Rename 'encrypt()' to 'yescrypt_sha256_cipher()'.

besser82:
Coverity Scan erroneously detects this function as a risky cryptographic function and generates several different false positive issues about it.

Beside that we shouldn't use function names of well known system functions internally to avoid incidental clashes.

***

@solardiz @vt-alt What do you think?


solardiz:
I'm OK with this. I might or might not come up with a different name for this function later, but since it's internal to this source file now, we can make the change as you propose now. Thanks!

besser82:
Allrighty!  Will merge this as soon as TravisCI goes green.

PR #50, opened 2018-11-01
The 'key' parameter for 'des_set_key()' must be at least 8 bytes long.

besser82:
Coverity scan complains that tainted data can be stored in 'rawkey1'.  Since there are excactly 8 bytes of data accessed, we should enforce this as a minimum parameter length.

### DIFF
--- a/alg-des.c
+++ b/alg-des.c
@@ -71,13 +71,14 @@ static const uint8_t key_shifts[16] =
 };
 
 void
-des_set_key (struct des_ctx *restrict ctx, const unsigned char *key)
+des_set_key (struct des_ctx *restrict ctx,
+             const unsigned char key[MIN_SIZE(8)])
 {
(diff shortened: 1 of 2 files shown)


besser82:
The description of the issue says 'Untrusted array index read'.  To me it sounds like it is concerned about the code may read out-of-bounds…

besser82:
@zackw I'll submit this PR to Coverity Scan before merging it.  Let's see whether it changes that problem…

solardiz:
Regarding the DES stuff, I previously (in e-mail a year ago) recommended switching to the FreeSec fork from musl (which I had put some effort into improving in multiple ways - lower memory needs, mostly glibc UFC-crypt compatible handling of invalid salts). I think this hasn't been done yet, or has it? While I don't mind addressing the individual Coverity issues in the current code, perhaps that will have to be re-done after the switch.

PR #51, opened 2018-11-01
Optimization to remove logically dead code in most configurations.

besser82:
Coverity Scan complains about some code path cannot be reached in most configurations.  Thus we should change the logic of the conditional to avoid it.


besser82:
Changed the comment.  Will merge when TravisCI goes green.

Issue #30, opened 2018-08-30
Issues from high-level review of yescrypt core code


zackw:
I looked over the yescrypt core code, mostly for integration problems with the libxcrypt environment.  I didn’t do a thorough review; this code is already mature, and I’m not qualified to review cryptographic primitives anyway.  I have some immediate concerns regarding the code in `yescrypt-platform.c`, but nothing else urgent.  There are also a number of long-term changes that we might want to make, mostly in the direction of reducing code duplication between this hashing method and others.

I would like to ask @solardiz a number of questions in his role as upstream yescrypt maintainer; all direct questions below are aimed at him.  I’m filing this as a bug on libxcrypt so I don’t lose it, but I’m happy to take the conversation to some upstream forum if you tell me where to go.

## General commentary

I see `_OPENMP` ifdefs in `yescrypt-opt.c`.  Does that actually work to the point where it would be worth the trouble for us to add a `--with-openmp` switch to the configure script and wiring it all up? Does this provide a useful performance gain for authentication uses, or is it just good for cracking?

I appreciate the detailed documentation comments on all the API functions in `yescrypt.h`.  It would be nice to have complete documentation of all the flag bits as well (it says “refer to the description of `yescrypt_kdf`,” but that only covers `YESCRYPT_WORM` vs `YESCRYPT_RW`).

Down the road, I think it would make sense for libxcrypt to expose more of the tunability that is available via the `yescrypt_*` functions, by adding a new version of the `crypt_gensalt` interface that can take more parameters.  It might even be desirable to bring back ROMs and other persistent state beyond what `struct crypt_data` can handle.  I haven’t thought much at all about what that would entail API-wise, but because of that, I am fine with keeping around code that is currently unreachable.

## Immediate concerns

These all relate to `yescrypt-platform.c`.

Nothing is arranging for `HAVE_POSIX_MEMALIGN` to be defined or not defined as appropriate.  This is fixable on our end, since it’s the standard macro that autoconf would define if asked to check for posix_memalign, but I would like to ask where it came from in the first place—I don’t see any machinery in Openwall’s yescrypt repo to set it.

[...]

PR #52, opened 2018-11-01
Drop code under #if HAVE_POSIX_MEMALIGN.

zackw:
alg-yescrypt-platform.c contains code to use posix_memalign if mmap is
unavailable, but it’s not properly wired up to autoconf, so it never
actually gets used.  I don’t think it’s worth keeping, because any
system that bothers to supply posix_memalign probably also has mmap.

I'd like @solardiz to comment on this change.

(This addresses one of the issues described in #30.)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -21,11 +21,6 @@ It was last updated 20 October 2018.
   * Make sure the symbol versioning macros work with all of the
     compilers that anyone needs (they use GCC extensions that clang
     also supports).
-  * `alg-yescrypt-common.c` contains #ifdeffage for use of
-    `posix_memalign` that isn’t wired up to autoconf
-    * possibly this should just be removed; does anything have
-      `posix_memalign` but not `mmap`? it’s only a compile-time
-      fallback, not a runtime fallback
 
 * Hardening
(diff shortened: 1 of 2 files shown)


solardiz:
I support this change and am likely to make the same change upstream.

However, I disagree with part of the reasoning that "any system that bothers to supply posix_memalign probably also has mmap" - I guess Windows is a notable exception here, it just might not be relevant to libxcrypt. My own reasoning is that we maintain alternative source code (and struct fields to support it) anyway and it isn't slower - it's just some trivial extra calculations. We can accept this trivial runtime complication on systems that have `posix_memalign` in order to simplify the source code and improve the coverage of its testing.

zackw:
I think I understand.  Would you describe this as a better rationale for the change?

> It is not worth attempting to use posix_memalign when mmap is not available, because the existing fallback to malloc + manual alignment is good enough for that situation.

(N.B. Current versions of Windows offer huge page allocations via VirtualAlloc, that might be worth looking into.)

solardiz:
Yes, that wording for the rationale sounds better to me. And thanks for the info on Windows VirtualAlloc.

zackw:
Rebased and pushed with the edited rationale.

PR #53, opened 2018-11-01
Fix explicit null derefences.

besser82:
Coverity complains it may happen under some, although very unlikely, conditions.

***

@solardiz  What do you think?

### DIFF
--- a/alg-yescrypt-opt.c
+++ b/alg-yescrypt-opt.c
@@ -1086,9 +1086,17 @@ static void smix(uint8_t *B, size_t r, uint32_t N, uint32_t p, uint32_t t,
 				HMAC_SHA256_Buf(Bp + (128 * r - 64), 64,
 				    passwd, 32, passwd);
 		}
-		smix1(Bp, r, Np, flags, Vp, NROM, VROM, XYp, ctx_i);
-		smix2(Bp, r, p2floor(Np), Nloop_rw, flags, Vp,
-		    NROM, VROM, XYp, ctx_i);
+		if (ctx_i) {
+			smix1(Bp, r, Np, flags, Vp, NROM, VROM,
+			    XYp, ctx_i);
+			smix2(Bp, r, p2floor(Np), Nloop_rw, flags, Vp,
+			    NROM, VROM, XYp, ctx_i);
+		} else {
+			smix1(Bp, r, Np, flags, Vp, 0, NULL,
+			    XYp, NULL);
+			smix2(Bp, r, p2floor(Np), Nloop_rw, flags, Vp,
+			    0, NULL, XYp, NULL);
+		}
 	}
 
 	if (Nloop_all > Nloop_rw) {


zackw:
I don't understand this code well enough to review this change.  I tried myself to figure out why Coverity thought this could happen and got lost in the control flow.

I'm going to defer to @solardiz.

besser82:
@zackw Well it's tricky…

Coverity thinks `VROM` can be non-null independently from evaluting `flags & YESCRYPT_RW` to `TRUE`.  Thus it complains…

solardiz:
What exactly does Coverity say about which lines? I understand the current logic might be too complicated to follow, but I'd rather come up with a "fix" that doesn't complicate the logic even further. And the title "Fix explicit null derefences." almost certainly doesn't apply.

zackw:
The problem is in how [https://github.com/besser82/libxcrypt/blob/develop/alg-yescrypt-opt.c#L1075-L1091](smix calls smix1 and smix2).  When `flags & YESCRYPT_RW` is not true, the final `pwxform_ctx_t *ctx_i` argument to smix1 and smix2 will be NULL.  Coverity thinks that they will then dereference it even though it's NULL.  I think this is not actually possible but the conditions involved are complicated enough that I'm not 100% sure myself.

besser82:
@solardiz That's the output from Coverity:

```
assign_zero         alg-yescrypt-opt.c:1075
var_deref_model     alg-yescrypt-opt.c:1090
deref_parm_in_call  alg-yescrypt-opt.c:956
deref_parm          alg-yescrypt-opt.c:737
```

besser82:
@solardiz And wit ha bit more context:

```
alg-yescrypt-opt.c:1075:  assign_zero: Assigning: ctx_i = NULL
alg-yescrypt-opt.c:1090:  var_deref_model: Passing null pointer ctx_i to smix2, which dereferences it.
alg-yescrypt-opt.c: 956:  deref_parm_in_call: Function blockmix_xor_save dereferences ctx.
alg-yescrypt-opt.c: 737:  deref_parm: Directly dereferencing parameter ctx.
```

solardiz:
`ctx` is set to non-NULL only if `YESCRYPT_RW` was set, and is only used in the same case. A subtle sub-case is `ctx` still being set to non-NULL if `YESCRYPT_RW` is later reset for the second `smix2` call with `VROM` set, which may also trigger use of `ctx`. Thus, the code looks correct to me as it is, and in case this logic is ever somehow broken I'd actually prefer a crash on a NULL dereference over having this swept under the rug with an extra (currently redundant) check. That said, I'll think of how to possibly simplify it and make Coverity happy.

solardiz:
I'd be surprised if the change currently proposed in this PR would make Coverity happy. I see no reason why it would, short of a possible Coverity bug.

besser82:
Closing here, as there has been a proposal for a better solution in https://github.com/besser82/libxcrypt/pull/56.

PR #55, opened 2018-11-01
Move helper functions into 'crypt-common.c'.

besser82:
Inlining these functions into exery compilation unit take time and wastes space in the library.  Additonally some compilers complain about unused inline functions.


besser82:
Rebase merged on cli.

PR #56, opened 2018-11-02
yescrypt: split smix1 and smix2 into functions for each subcase.

zackw:
smix’s subroutines smix1 and smix2 both do some setup work, then perform one of several different operations depending on the precise combination of arguments they were passed, and then some teardown work.  Some combinations of arguments are not possible but would cause crashes if they were possible.  Static analysis has trouble proving that these combinations are not possible.

Split up both subroutines into separate functions for each of the possible operations (smix1_rom, smix1_rw, smix1_worm; smix2_rom, smix2_romro, smix2_rw, smix2_rwro, smix2_worm).  The NROM, VROM, and ctx arguments are only supplied to the subroutines that will actually use them, and flags is supplied to none of them.  The shared setup and teardown work is factored out to helper routines smix_init and smix_final. All of the “which case are we in?” logic is consolidated into smix itself.

This should be more legible to both static analysis programs and human understanding.  Specifically, I believe it will resolve the Coverity defect report that ctx could be dereferenced while it was NULL.  However, I don't think libxcrypt's test suite covers all of the cases that are affected (since we don't ever use ROMs or p>1) and it's a significant enough change that I don't think we should merge it without yescrypt upstream agreeing to take it as well.

I would like @solardiz to review this change but I do not seem to be able to officially request that through the Github interface.

### DIFF
--- a/alg-yescrypt-opt.c
+++ b/alg-yescrypt-opt.c
@@ -797,21 +797,13 @@ static inline uint32_t integerify(const salsa20_blk_t *B, size_t r)
 }
 
 /**
- * smix1(B, r, N, flags, V, NROM, VROM, XY, ctx):
- * Compute first loop of B = SMix_r(B, N).  The input B must be 128r bytes in
- * length; the temporary storage V must be 128rN bytes in length; the temporary
- * storage XY must be 128r+64 bytes in length.  N must be even and at least 4.
- * The array V must be aligned to a multiple of 64 bytes, and arrays B and XY
- * to a multiple of at least 16 bytes.
+ * smix_init(B, r, X, Y)
+ * Perform the initial round of decodes and salsa20 shuffles used by both
+ * smix1 and smix2.
  */
-static void smix1(uint8_t *B, size_t r, uint32_t N, yescrypt_flags_t flags,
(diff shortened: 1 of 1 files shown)


solardiz:
Thanks, but this appears to add 167 lines of code and probably makes further maintenance harder and has unclear to me effect on "human understanding". I think this shouldn't be merged as-is, but I'd like to revisit these changes and maybe cherry-pick some ideas when I try to simplify upstream code and make it friendlier to static analysis. I might reuse the smix_init and smix_final concept - in fact, I've already been thinking that it might be simpler to do those things at top smix rather than smix1/smix2 level.

zackw:
OK, understood.  I'm not going to pursue this any further myself.  Please
let us know if and when you have a change that you're happy with that might
address the issue.

besser82:
@zackw @solardiz For the time until this is fixed in some way, I've added special annotations for Coverity Scan in https://github.com/besser82/libxcrypt/commit/42a22323a0267da38fffd873246e08950423753c.

solardiz:
@besser82 The annotations for Coverity are OK with me for now, thanks!

Issue #54, opened 2018-11-01
Consider introducing limits on resource usage by maybe-rogue hash encodings


solardiz:
Historically, hashes were fixed-cost and thus OK for semi-trusted users to be able to specify directly e.g. in Apache httpd `.htpasswd` files. With tunable-cost hashes, this changes - a semi-trusted user could DoS the service for its other users, and if no adequate resource limits are configured on the service then also DoS the system. musl chose to impose some hard-coded sanity limits on those hashes. With libxcrypt's support for hashes tunable not only for time, but also for memory (scrypt, yescrypt), this may be even more of an issue.

I am undecided on whether we should merely document the issue or also impose limits, and if so whether they should be hard-coded or configurable, and if so compile- or run-time configurable. To avoid duplicate parsing, some logic might need to be added to upstream yescrypt tree and made use of by libxcrypt.

At least we should make an informed decision. Hence opening this issue.

zackw:
The `crypt.conf` format I proposed in #26 could easily be extended to support run-time configurable upper limits.

I think we should certainly document the issue, and I think there's a strong case for doing more than that, but I am not sure I know what that is. Do you have links to the discussion that happened when musl added limits?

solardiz:
Here are two relevant messages/threads:

https://www.openwall.com/lists/musl/2012/08/08/20
https://www.openwall.com/lists/oss-security/2013/06/12/1

zackw:
I just revised the [crypt.conf spec](https://github.com/besser82/libxcrypt/pull/26/commits/159d89c2e9a6fef66c5d5f37f7102407e4f49be8#diff-74d4bc28af6647750c34c24a894dd573) so that there are now `defcost=` and `maxcost=` keyval arguments for each hashing method.  We'd need to add more knobs if we ever allow tuning memory separately from time parameters, but I think this will do for the time being.

Unresolved questions:

 * Does the documentation I wrote explain the issue clearly enough?
 * What should the default upper time limit used by `crypt-tune-costs` be?  One second? Ten?
 * Should there be a `mincost=` separate from `defcost`?  Right now, the documented behavior is that you _can_ specify costs lower than `defcost` when calling `crypt_gensalt` but the result will be flagged as "too cheap" by `crypt_checksalt`.  It might make sense to allow these to be controlled separately; it might also make sense to disallow costs lower than `defcost` in `crypt_gensalt`, I am undecided.

solardiz:
[...]

PR #57, opened 2018-11-07
Implement crypt_checksalt.

besser82:
This function can be used by portable users of libxcrypt to check whether the desired hash method is supported.

There are future plans to extend the crypt_checksalt function to check the given setting string against the configuration in '/etc/crypt.conf' and report whether the hashing method and parameters it specifies are acceptable.

***

@ldv-alt I'd like to get you feedback, too, as this the base for a future implementation of runtime checks in `linux-pam`.


solardiz:
I think this would make sense if we implemented it in a way not wasting time and memory on a test hash computation. With the current straightforward implementation, it only makes sense to me as an initial stub intended to be reimplemented later, because that straightforward implementation is also possible and simple to do directly in the caller without us introducing this API.

Also, maybe (just maybe) we could address #54 in the same API - e.g., have it return estimated would-be time and memory needs, or have it fail if supplied limits would be exceeded. Just something to consider.

zackw:
There is already a manpage for this API on the crypt.conf branch that you can grab. I would prefer that you use the constants documented in that manpage instead of the constants you proposed.

For now, since we currently don't have any way to make hash methods supported for old hashes but not new ones, you could have it just return 0 when `get_hashfn` returns a non-null value, and `CRYPT_SALT_INVALID` when it doesn't.  That would avoid having to do a dummy encryption, which I was also concerned about.

Please do write a proper test.

besser82:
@zackw All changes are made in the rebased commit.

About the manpage:  We possibly need to revise its wording to match the current state of implementation…

zackw:
The code looks good.  I would suggest that you have the test case call `crypt_gensalt` for each prefix, `crypt_checksalt` on the result, `crypt` with a stock passphrase, and then` crypt_checksalt` again.  So that we're also testing `crypt_gensalt` for complete setting strings and complete hashed passphrases.

Regarding the documentation, I suggest changing the first paragraph of DESCRIPTION to say that the function checks against "the system configuration" instead of "the configuration in /etc/crypt.conf", dropping the cross-reference to crypt.conf.5 in SEE ALSO, and adding a BUGS section that says that the current implementation will only ever return 0 or CRYPT_SALT_INVALID because full configurability is not yet implemented.  Also, CRYPT_SALT_OK should be mentioned in this file.  (It's important to be clear that CRYPT_SALT_OK is guaranteed to be zero, though.)

besser82:
@zackw Everything should be addressed in the rebased commit.

PR #59, opened 2018-11-09
Configure the value of 'CRYPT_GENSALT_IMPLEMENTS_DEFAULT_PREFIX'.

besser82:
The constant must only be defined to '1', if the default hashing method is enabled in the configuration.
Otherwise it must be defined to '0'.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -116,6 +116,10 @@ crypt.h: crypt.h.in crypt-hashes.h crypt-symbol-vers.h gen-crypt-h.awk config.h
 	$(AM_V_GEN)LC_ALL=C $(AWK) \
 	  -f $(srcdir)/gen-crypt-h.awk config.h $(builddir)/crypt.h.in \
 	  > crypt.h.T
+	$(AM_V_GEN)LC_ALL=C \
+	$(GREP) -q "#define HASH_ALGORITHM_DEFAULT" crypt-hashes.h && \
+	$(SED) -i -e "s/@DEFAULT_PREFIX_ENABLED@/1/g" crypt.h.T || \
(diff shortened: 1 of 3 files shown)


besser82:
@zackw We need to document this?

zackw:
It makes sense, but I wonder if we should instead make sure that there _is_ always a default hashing method.

besser82:
> It makes sense, but I wonder if we should instead make sure that there _is_ always a default hashing method.

I thought about that, too… But how do we determine which prefix (e.g. the strongest from the given selection) to be the default?

zackw:
Strongest among those enabled at configure time is what I was thinking.  `hashes.lst` is _almost_ in descending order of cryptographic strength; the only exceptions are that the relative strengths of md5crypt, sunmd5, and sha1crypt are debatable (this we could easily tweak), and nt is arguably _weaker_ than the DES-based hashes because it doesn't have any salt (this we can't change easily, because nt has a prefix and descrypt doesn't).  These all affect hashes that are not STRONG.  If someone turns off all of the STRONG hashes, well, presumably they know what they're doing, but refusing to have crypt_gensalt generate settings for a non-STRONG hash without an explicit request seems like a useful additional safety feature.

So, here's my suggestion: Land this change as-is, but as a follow-up, get rid of the DEFAULT annotation in hashes.lst.  The awk scripts select as default the first entry in hashes.lst that is enabled *provided that it is a STRONG hash*.  If none of the STRONG hashes are enabled, there is no default.

besser82:
@zackw Do you want to deal with `gen-hashes.awk`?  Just removing the `DEFAULT` annotation doesn't work…

zackw:
Yeah, I'll look at it this evening.

besser82:
That would be great!  I'm planning to release v4.3.0 during the weekend.  ;)

PR #60, opened 2018-11-11
Remove the gensalt fuction for the bcrypt x variant ($2x$).

besser82:
This prefix was never intended for use when hashing new passphrases.
The only use case originally intended was to manually edit '$2a$' to '$2x$' in passphrase hashes to knowingly take the risk yet enable users to continue to log in when upgrading systems with buggy bcrypt implementations to fixed versions.

There was never an intent to be able to generate new setting strings with that prefix.  The original implementation of the bcrypt gensalt function doesn't allow the use of the prefix '$2x'.
Thus libxcrypt must not, either.


crypt-bcrypt.c
```diff
@@ -1028,11 +1029,14 @@ crypt_bcrypt_x_rn (const char *phrase, size_t ARG_UNUSED (phr_size),
...
-                     uint8_t *output, size_t o_size)
+gensalt_bcrypt_x_rn (ARG_UNUSED(unsigned long count),
+                     ARG_UNUSED(const uint8_t *rbytes),
+                     ARG_UNUSED(size_t nrbytes),
+                     ARG_UNUSED(uint8_t *output),
+                     ARG_UNUSED(size_t o_size))
 {
-  BF_gensalt ('x', count, rbytes, nrbytes, output, o_size);
+  errno = EINVAL;
+  return;
```
solardiz: Do we need a `gensalt_bcrypt_x_rn` at all? Ideally, we shouldn't.

besser82: It is easier to keep this function, since the function is excpeted to be present by all the autogen magic.  Thus I've made it to set `errno` to `EINVAL` and explicitly return.

besser82:
> The (subtype != 'a' && subtype != 'b' && subtype != 'x' && subtype != 'y') check in BF_gensalt should have the subtype != 'x' portion dropped. That would revert this code back to what it is in crypt_blowfish.

@solardiz Fixed in rebased commit.  I've also added a comment in `gensalt_bcrypt_x_rn()` stating that this prefix must not be used for computing new hashes.

besser82:
@zackw Are you fine with this change?

test-gensalt-extradata.c
```diff
@@ -85,6 +82,7 @@ static const struct testcase testcases[] =
 #if INCLUDE_gost_yescrypt
   { "$gy$", 7, 11, },
 #endif
+  { "", 0, 0, }
```
zackw: Why is this change necessary?

test-gensalt-extradata.c
```diff
@@ -149,7 +147,7 @@ main(void)
   char sref[6][CRYPT_GENSALT_OUTPUT_SIZE];
   char stst[CRYPT_GENSALT_OUTPUT_SIZE];
 
-  for (size_t i = 0; i < ARRAY_SIZE(testcases); i++)
+  for (int i = 0; i < (int) ARRAY_SIZE(testcases) - 1; i++)
```
zackw: This also seems unnecessary...

test-gensalt-extradata.c
```diff
@@ -184,7 +182,7 @@ main(void)
         }
       else
         {
-          printf ("ERROR: %zu/%s: inappropriate count1=%lu count2=%lu\n",
+          printf ("ERROR: %d/%s: inappropriate count1=%lu count2=%lu\n",
```
zackw: ... and this.

test-gensalt-extradata.c
besser82: If we configure with `--enable-hashes=bcrypt_x` explicitly, the `testcases` array needs at least one (and in this case the terminating) element.

test-gensalt-extradata.c
besser82: As commented above in a certain corner case we need a terminating element in the array.  With this change I make sure, we do not iterate the last element.

test-gensalt-extradata.c
zackw: What's wrong with the testcases array having zero elements?

test-gensalt-extradata.c
besser82: Needed as explained previously.

besser82:
> Apart from the apparently-unrelated changes to the main loop in test-gensalt-extradata, this looks good to me.

@zackw Do you agree with my explaination?

test-gensalt-extradata.c
zackw: Oh, right, ISO C forbids zero-length arrays.  In that case, please change `const char prefix[8];` in `struct testcase` to `const char *prefix;` so that we can use `{ 0, 0, 0 }` for the terminator (avoiding potential confusion with the `""` prefix used for descrypt) and make the main loop condition be `for (size_t i = 0; testcases[i].prefix; i++)` and then `i` can remain a `size_t`.

besser82: Changed in rebased commit.

besser82:
@zackw Waiting for Travis to finish.  I will merge then and push `v4.3.2` bugfix release.

PR #62, opened 2018-11-12
[RFC] [DON'T MERGE YET] Make crypt and crypt_gensalt use thread-local output buffers.

zackw:
This change makes crypt and crypt_gensalt as thread-safe as they can be without changing their interfaces.  Solaris already made this change, and it’s being discussed by glibc (with suggestion that it
should be pushed upstream to the C and POSIX standards committees): https://sourceware.org/ml/libc-alpha/2018-10/msg00437.html

Portable programs should still use the r-variants, though, because this is not a guaranteed feature, it doesn’t make them not clobber their output buffers on a second call, and the tradeoff is a sizeable
memory leak (CRYPT_GENSALT_OUTPUT_SIZE or sizeof(crypt_data)) for each thread that calls either function.  (Thanks ever so much, C committee, for adding thread-local variables without destructors for them.) (I’m still investigating whether there’s some GNU extension we could use to avoid the memory leak at least on Linux.)  (We could use pthread_getspecific, but then we’d have to link libcrypt with libpthread, and this doesn’t seem like enough reason to do that.)

I'd like comments on this, but please don't merge it yet; the memory leak is a scary cost and I would like to be sure there's really no way to avoid it before merging this as-is.

(On the up side, this change shrinks libcrypt.so's .bss segment from 33160 to 192 bytes, which is a nice savings for programs that don't use these functions.)


crypt-gensalt-static.c
```diff
@@ -15,20 +15,29 @@
...
 #include "crypt-port.h"
+#include <stdlib.h>
 
 /* The functions that use global state objects are isolated in their
    own files so that a statically-linked program that doesn't use them
    will not have the state objects in its data segment.  */
 
 #if INCLUDE_crypt_gensalt
+
+static THREAD_LOCAL char *obuf;
```
besser82: Shouldn't we explicitly initialize the pointer with a `NULL` value?

crypt-static.c
```diff
@@ -15,17 +15,43 @@
...
 #include "crypt-port.h"
+#include <stdlib.h>
 
 /* The functions that use global state objects are isolated in their
    own files so that a statically-linked program that doesn't use them
    will not have the state objects in its data segment.  */
 
 #if INCLUDE_crypt || INCLUDE_fcrypt
+
+static THREAD_LOCAL struct crypt_data *nr_crypt_ctx;
```
besser82: Same about explicit initialization with `NULL` value…

crypt-gensalt-static.c
zackw: Technically not necessary (static pointers are auto-initialized to NULL, thread-local or not) but I suppose it would be clearer.

besser82:
Mhh…  Is there any chance to abuse `__cxa_thread_atexit_impl` to `free()` the pointers?

besser82:
@zackw I've just pushed a possible abuse of `__cxa_thread_atexit_impl` to free the pointers at thread termination.  The only drawback:  Only exists in glibc > 2.18, no other systems supported.

rfc1036:
I am not sure that I understand the point of this change.

Current situation:
- single threaded programs already work fine
- multithreaded programs probably already use the reentrant API

After this patch:
- single threaded programs will leak memory unless they switch to the reentrant API
- no obvious need for the changed API

So this change would force long-lived daemons to use a different API while providing no obvious real world benefits.

zackw:
@besser82 I believe I know how to do that in a more universally supported way - the only wrinkle being that one file will need to be compiled as C++, which is enough of a wrinkle that I'm not gonna get to it this week, probably.

@rfc1036 The primary motivation for this change is to match what Solaris already does, and what has been proposed to be done in glibc with all the _other_ APIs that have static internal buffers.  See the thread I linked to at the beginning.  A secondary motivation is to reduce the size of libcrypt.so's BSS. That said, I'm certainly open to the possibility that it's not worth bothering with.

besser82:
Resolved merge conflicts.

besser82:
Rebased on recent `develop`, resolved merge conflicts.

solardiz:
I think this is a questionable feature, primarily because portable programs must not rely on it anyway, and then there's risk of someone porting system-specific code e.g. from Fedora to another system and not realizing they need to revise the uses of crypt*(). There's already such risk with ports from Solaris, but I think that's a poor excuse for making the matter worse. I wouldn't merge this unless and until glibc actually makes similar changes across the board.

besser82:
> > Wouldn't that imply we need to link libstdc++.so
> 
> No. The short version is that the particular C++ feature needed for this (a `static thread_local` object, declared at function scope with constructor and destructor) winds up calling `__cxa_thread_atexit_impl` via a shim that GCC helpfully provides in a static-only library called `libsupc++`. As long as we don't use any _other_ C++ features (e.g. continuing to do the large allocation with `malloc` instead of `new`), we don't need the full C++ runtime. And the advantage of doing it this way is we don't have to worry about whether `__cxa_thread_atexit_impl` is actually available; the shim takes care of that for us.

Unfortunately `libsupc++` isn't portable.  Systems with Clang/LLVM only are not guaranteed to ship an equivalent shim-a-like library, tho some of them may ship an externally provided `libcxxrt`.

Given this fact and the very likely need for `pthreads` when implementing `crypt.conf`, I'm biased towards to start using `pthreads` here as well.

besser82:
Anyways, looking at the [code from (Open)Solaris](https://github.com/illumos/illumos-gate/blob/cb6207858a9fcc2feaee22e626912fba281ac969/usr/src/lib/libc/port/gen/crypt.c) their implementation of thread-safety in the `crypt` function isn't more then a `mutex` around the internal buffer of DES `setkey` and `encrypt`.  They do not use TLS or any other advanced things.

test-crypt-mt.c
```diff
@@ -0,0 +1,189 @@
...
+      printf ("ERROR: parent: crypt failed (\"%s\"): %s\n",
+              hash, strerror (errno));
+      return 1;
+    }
+  else
+    printf ("ok: parent: crypt = %s\n", hash);
+  if (strlen (hash) + 1 > sizeof save_hash)
+    {
+      printf ("ERROR: crypt output is too long (%zu > %zu)\n",
+              strlen (hash) + 1, sizeof save_hash);
```
besser82: Codacy Issue found: [Does not handle strings that are not \0-terminated; if given one it may perform an over-read (it could cause a crash if unprotected) (CWE-126).  ](https://app.codacy.com/manual/besser82/libxcrypt/pullRequest?prid=5024327)

besser82:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 7
- Added 4
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/besser82/libxcrypt/pullRequest?prid=5024327&bid=16558570)

test-crypt-mt.c
```diff
@@ -0,0 +1,189 @@
...
+
+  setting = crypt_gensalt (HASH_PREFIX, 0, rbytes1, 64);
+  if (!setting)
+    {
+      printf ("ERROR: parent: crypt_gensalt failed: %s\n", strerror (errno));
+      return 1;
+    }
+  else
+    printf ("ok: parent: crypt_gensalt = %s\n", setting);
+  if (strlen (setting) + 1 > sizeof save_setting)
```
besser82: Codacy Issue found: [Does not handle strings that are not \0-terminated; if given one it may perform an over-read (it could cause a crash if unprotected) (CWE-126).  ](https://app.codacy.com/manual/besser82/libxcrypt/pullRequest?prid=5024327)

test-crypt-mt.c
```diff
@@ -0,0 +1,189 @@
...
+    {
+      printf ("ERROR: crypt_gensalt output is too long (%zu > %zu)\n",
+              strlen (setting) + 1, sizeof save_setting);
```
besser82: Codacy Issue found: [Does not handle strings that are not \0-terminated; if given one it may perform an over-read (it could cause a crash if unprotected) (CWE-126).  ](https://app.codacy.com/manual/besser82/libxcrypt/pullRequest?prid=5024327)

test-crypt-mt.c
```diff
@@ -0,0 +1,189 @@
...
+    }
+  else if (hash[0] == '*' || hash[0] == '\0')
+    {
+      printf ("ERROR: parent: crypt failed (\"%s\"): %s\n",
+              hash, strerror (errno));
+      return 1;
+    }
+  else
+    printf ("ok: parent: crypt = %s\n", hash);
+  if (strlen (hash) + 1 > sizeof save_hash)
```
besser82: Codacy Issue found: [Does not handle strings that are not \0-terminated; if given one it may perform an over-read (it could cause a crash if unprotected) (CWE-126).  ](https://app.codacy.com/manual/besser82/libxcrypt/pullRequest?prid=5024327)

solardiz:
> Anyways, looking at the [code from (Open)Solaris](https://github.com/illumos/illumos-gate/blob/cb6207858a9fcc2feaee22e626912fba281ac969/usr/src/lib/libc/port/gen/crypt.c) their implementation of thread-safety in the `crypt` function isn't more then a `mutex` around the internal buffer of DES `setkey` and `encrypt`. They do not use TLS or any other advanced things.

They do use `tsdalloc` for `crypt` output buffer: https://github.com/illumos/illumos-gate/blob/cb6207858a9fcc2feaee22e626912fba281ac969/usr/src/lib/libc/port/gen/crypt.c#L154
and it has some thread magic inside: https://github.com/illumos/illumos-gate/blob/cb6207858a9fcc2feaee22e626912fba281ac969/usr/src/lib/libc/port/gen/tsdalloc.c#L66
(linking here to this same old revision you had referred to in your comment in 2019)

PR #65, opened 2018-11-14
Add 'fedora' to the groups of hashing methods.

besser82:
--enable-hashes now supports 'fedora' as a group of hashing methods.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,7 @@ Please send bug reports, questions and suggestions to
 <https://github.com/besser82/libxcrypt/issues>.
 
 Version 4.3.4
+* --enable-hashes now supports 'fedora' as a group of hashing methods.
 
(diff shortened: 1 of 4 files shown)


besser82:
> LGTM _except_ shouldn't a bunch more hashes be tagged FEDORA? Probably at least all of the GLIBC ones? We have been trying to make the hash groups self contained.

@zackw Following this logic, we would need to add all hashes from the `GLIBC` group to `ALT`, `OWL`, `SUSE`…  From my perspective the `GLIBC` group is present, so we do not need to list any linux distribution per hash explicitly.

The reason for listing `*BSD` and others to single hashes, is they all use their own implementation of libc and thus a custom libcrypt with several differences between which hashes are actually supported.

Do you agree here?

PR #67, opened 2018-11-17
Add the configure option --enable-unity-build.

besser82:
Unity builds are a technique for cutting down build times.  The way it works is relatively straightforward.

Normally we would run the compiler three times, once for each source file individually; in a unity build we instead compile all these sources within a single unit.

The simplest approach is to create a new source file that looks like this:

    #include<src1.c>
    #include<src2.c>
    #include<src3.c>
    …

This generated source file is then compiled rather than the individual source files one by one.

The exact speedup depends on the code base, of course, but it is not uncommon to obtain more than 50% speedup in compile times.  This happens even though the Unity build uses only one CPU whereas individual compiles can be run in parallel.

Unity builds can also lead to faster code, because the compiler can do more aggressive optimizations (e.g. inlining).

The downside is that incremental builds are as slow as full rebuilds (because that is what they are).  Unity compiles also use more memory, which may become an issue in certain scenarios.


solardiz:
I think this is a controversial feature that isn't worth the risk (of having missed a clash, of having introduced a bug along with the code edits, of enabling an unsafe optimization, even of triggering a compiler bug) and the deviation from upstream code revisions (have to patch the clashes - not only now, but also when merging changes later). Also, having more ways to build the project means that each will receive less testing, and there will be more bugs the project will be responsible for.

I think this is only potentially useful as an extra stress-test, not as a feature for actual use. But then if we don't permanently patch the clashes and maintain such changes, even the stress-test will be a one-time thing we can run now, not something to maintain and re-run. So I see very little use in this.

Are libxcrypt build times too high? If so, why? Maybe there's another way to address the problem.

Also, I doubt the reduction in build times is reliable - I would expect that the larger translation units would sometimes take longer to build/optimize than their components would take separately.

zackw:
I agree with @solardiz that this doesn't seem worth the complexity it adds.  Our build matrix is already awkwardly large, and we already have a bunch of non-obvious (and undocumented) coding constraints that will only get caught by the compiler in one configuration.

I would like to suggest some alternative angles of attack that should accomplish at least some of the goals you were striving for:

* Generating a precompiled header from crypt-port.h and everything that it includes.  (I have no idea how difficult it will be to make this cooperate with libtool and/or MacOS universal builds.)
* Reducing the number of system headers included by crypt-port.h.  Of course, lots of .c files may wind up including them directly if we do that.
* Figuring out some way to make link-time optimization play nice with symbol versioning.

besser82:
@zackw What about merging the single commit https://github.com/besser82/libxcrypt/pull/67/commits/dbd644af9a7f80e8ce1b195b7c74339f701d889b into develop?

zackw:
@besser82 
> What about merging the single commit dbd644a into develop?
I like that idea.  I'll respond in more detail in #69.

PR #68, opened 2018-11-18
Implement the crypt_preferred_method function.

besser82:
crypt_preferred_method(3) is a convenience function to get the prefix of the preferred hash method.  If a preferred method is available, it is the same as the one also used by the crypt_gensalt functions, if their given prefix parameter is NULL.

If no preferred hash method is supported, the return value of this function is NULL.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,6 +81,7 @@
 /test-gensalt-extradata
 /test-getrandom-fallbacks
 /test-getrandom-interface
+/test-preferred-method
 /test-short-outbuf
 
 # backup-files
--- a/Makefile.am
+++ b/Makefile.am
@@ -24,7 +24,7 @@ EXTRA_DIST = \
 notrans_dist_man3_MANS = \
 	crypt.3 crypt_r.3 crypt_ra.3 crypt_rn.3 \
 	crypt_checksalt.3 crypt_gensalt.3 crypt_gensalt_ra.3 \
-	crypt_gensalt_rn.3
+	crypt_gensalt_rn.3 crypt_preferred_method.3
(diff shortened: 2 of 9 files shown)


zackw:
This seems like a fine addition and I don't see any problems with the code.

besser82:
@zackw The documentation is fine, too?  Or does it need any changes / improvements?

zackw:
I'm not entirely happy with the bit that goes

> ```nroff
> If no preferred hash method is available it is NULL.  It
> .Em is
> safe to pass the string returned by
> .Nm crypt_preferred_method
> directly to
> .Nm crypt_gensalt .
> ```

because that makes it sound like there might be a thread-safety issue, even though there isn't.  But I haven't been able to think of a better phrasing and I don't think it should block the merge.  We can always improve the manpage later.

besser82:
> I'm not entirely happy with the bit that goes
> 
> > ```
> > If no preferred hash method is available it is NULL.  It
> > .Em is
> > safe to pass the string returned by
> > .Nm crypt_preferred_method
> > directly to
> > .Nm crypt_gensalt .
> > ```
> 
> because that makes it sound like there might be a thread-safety issue, even though there isn't. But I haven't been able to think of a better phrasing and I don't think it should block the merge. We can always improve the manpage later.
> 
> Otherwise the new docs look fine to me.

@zackw Maybe this sounds better?

> ```
> If no preferred hash method is available it is NULL.  It
> .Em is
> safe to pass the string returned by
> .Nm crypt_preferred_method
> directly to
> .Nm crypt_gensalt
> without prior string-sanitizing nor NULL-pointer checks.
> ```

PR #69, opened 2018-11-19
Move the base64 encoding table to crypt-common.

besser82:
Cherry-picked commit from #67.

***

@solardiz


crypt-common.h
```diff
@@ -0,0 +1,31 @@
...
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+/* Simple commonly used helper constants.  */
+
+#ifndef _CRYPT_COMMON_H
+#define _CRYPT_COMMON_H 1
+
+/* Table with characters for base64 transformation.  */
```
zackw: I would like this comment to be a little more detailed.  Something like

```c
/* The base-64 encoding table used by most hashing methods.
   (bcrypt uses a slightly different encoding.)  Size 65
   because it's used as a C string in a few places.  */
```

crypt-common.h
```diff
@@ -0,0 +1,31 @@
...
+const unsigned char ascii64[65];
+
+/* Same table gets used with other names in various places.  */
+#define b64t   ascii64
+#define itoa64 ascii64
```
zackw: You're doing this to minimize the number of changes we make to the `.c` files that are imported from elsewhere, right?  It looks like all of the places that use the name `b64t` want plain `char`, so you could eliminate all of the explicit casts that you introduced by having it be

``` c
#define b64t   ((const char *)ascii64)
```

solardiz:
OK with me. Would it possibly be cleaner for the array elements to be `char`, not `unsigned char`? The APIs are generally such that the returned hash encoding string is `char *`, and we know that all of the possible 64 chars fit in 7-bit ASCII.

If I have time next year or so, I might introduce generic cache-timing-safe encoding/decoding functions into yescrypt, which we would be able to use for other password hash methods as well. So for now staying with this minimal change and not trying to merge the encoding/decoding functions as well makes sense to me.

crypt-common.h
besser82: Fixed in rebased commit.

crypt-common.h
besser82: Likewise.

besser82:
> OK with me. Would it possibly be cleaner for the array elements to be `char`, not `unsigned char`? The APIs are generally such that the returned hash encoding string is `char *`, and we know that all of the possible 64 chars fit in 7-bit ASCII.

I intentionally used `unsigned char` here, as it reduces the amount of contextual changes needed to other files.

PR #70, opened 2018-11-19
Use prebuilt object files from libcrypt for linking the tests.

besser82:
Partially cherry-picked from #67.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,13 +4,15 @@
 # Author: Thorsten Kukuk <kukuk@suse.de>
 #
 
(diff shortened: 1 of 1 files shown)


zackw:
It occurs to me that this is arcane enough to deserve a comment in the Makefile.  Right above the definition of `COMMON_TEST_OBJECTS`:

``` make
# Refer to object files that are linked into libxcrypt with the
# qualified name, libcrypt_la-foobar.lo, to prevent them from
# being compiled a second time.
```

Issue #71, opened 2018-11-28
Consider building with `-fno-plt` when available


zackw:
This is a minor optimization that came to my attention earlier today and I'm filing this issue mostly so I don't forget about it.  (I'm not going to have any time at all for xcrypt work for the next several weeks.)

This is the GCC documentation for `-fno-plt`:

> Do not use the PLT for external function calls in position-independent code. Instead, load the callee address at call sites from the GOT and branch to it. This leads to more efficient code by eliminating PLT stubs and exposing GOT loads to optimizations. On architectures such as 32-bit x86 where PLT stubs expect the GOT pointer in a specific register, this gives more register allocation freedom to the compiler. Lazy binding requires use of the PLT; with -fno-plt all external symbols are resolved at load time.

Since we are already using `-Wl,-z,now` to disable lazy binding, the only downside (on x86-64) is that call instructions targeting C library functions are now six bytes each, instead of five bytes.  In very casual testing this increased the size of the text section by about 80 bytes total.  And completely eliminated the PLT section, so the overall code size change is effectively nil.  I was hoping it would reduce the overall number of relocations, but it didn't.

PR #72, opened 2018-12-02
Add -fno-plt to the compiler flags, if supported.  (#71)

besser82:
Do not use the PLT for external function calls in position-independent code.  Instead, load the callee address at call sites from the GOT and branch to it.  This leads to more efficient code by eliminating PLT stubs and exposing GOT loads to optimizations.

Since we are already using -Wl,-z,now to disable lazy binding, the only downside (on x86-64) is that call instructions targeting C library functions are now six bytes each, instead of five bytes.

Closes: #71


besser82:
@zackw The two functions are provided in your `zw_simple_warnings.m4`…

Shall I remove them there and include the original files instead?

zackw:
@besser82 Huh. I don't remember why I did that. Now that there's more than one use it probably makes sense to move toward keeping things the way the autoconf archive has them.  (zw_simple_warnings.m4 really is an improvement on ax_manywarnings.m4 IMO, but only the part that actually picks warnings to activate, not the infrastructure.)

zackw:
Since I'm still too busy with the day job to do any actual work on libxcrypt myself right now, you should do however much cleanup you feel is worth doing.

besser82:
Merged via https://github.com/besser82/libxcrypt/commit/4b4aaa2241aa8b8d52e78cbf486acb0d8cb1e67d.

PR #73, opened 2018-12-08
Fix the actually not happening explicit null dereference found by CovScan.

besser82:
This small change fixes the false positive explicit null dereference found by Coverity Scan in the smix{1,2} functions from alg-yescrypt-opt.c.

### DIFF
--- a/lib/alg-yescrypt-opt.c
+++ b/lib/alg-yescrypt-opt.c
@@ -85,6 +85,7 @@
 #include <xmmintrin.h>
 #endif
 
+#include <assert.h>
 #include <errno.h>
 #include <stdint.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


solardiz:
Well, this is relatively non-invasive, so is OK in that respect. However, it makes things worse: if with future code changes `ctx` does happen to be `NULL` where it must not be, with these changes we'll get incorrect computation results rather than a crash. I'd prefer a crash.

So I'd prefer that we keep the magic comments for Coverity.

But this is up to you and @zackw.

Thanks!

besser82:
@solardiz I've changed this another time to use `assert(ctx != NULL)`.  Are you fine with merging this solution?  @zackw What do you think?

solardiz:
@besser82 I think we're fixing a non-issue, but I am fine with these changes considering the code would crash anyway if `ctx` were NULL in any of those places.

There was the "Asserts considered harmful" thread started by Jeffrey Walton on oss-security in December and contunuing into January (and he had previously e-mailed me about that topic). Jeffrey advocates against asserts in library code. In our case, the asserts are harmless as it's supposed to be a "can't happen" condition within the library code itself and if it does happen then similar arguments to Jeffrey's could have been made regarding the NULL pointer dereference crashes. So the asserts don't make things worse. They also don't make things better, so I don't intend to include them upstream.

PR #75, opened 2019-01-08
Add --enable-obsolete-api-enosys configure option.

besser82:
If enabled, this option replaces the obsolete APIs (fcrypt, encrypt{,_r}, and setkey{,_r}) with POSIX compliant stubs setting "errno" to "ENOSYS", instead of actually performing any real operation.

The encrypt{,_r} function will - for security reasons - additionally overwrite the data-block argument with random data.

In case of the fcrypt function a unique failure-token (*0 or *1) will be returned; when failure-tokens have been disabled during build-time fcrypt will always return NULL.


besser82:
@zackw Are you still busy and/or travelling?

zackw:
@besser82 I have been on vacation for the past month but I'm back now.

I'll have some line notes for you in a short while, but I like the idea of this option.  It gives conservative distributions an option for turning off DES while preserving binary compatibility -- of course any program that _uses_ these APIs breaks either way, but runtime failure is maybe more tolerable.

I'm growing concerned by the sheer number of configure-time options we have, and the associated size of the test matrix, but this one does seem worth it.

NEWS
```diff
@@ -6,6 +6,15 @@ Please send bug reports, questions and suggestions to
 Version 4.4.3
 * Fix the value of SUNMD5_MAX_ROUNDS.
 * Add generated C++-guards to <xcrypt.h>.
+* Add --enable-obsolete-api-enosys configure option.
+  If enabled, this option replaces the obsolete APIs (fcrypt,
+  encrypt{,_r}, and setkey{,_r}) with POSIX compliant stubs setting
+  "errno" to "ENOSYS", instead of actually performing any real operation.
```
zackw: "... with stubs that set errno to ENOSYS and return without performing any real operations. This allows one to disable DES support while preserving POSIX compliance."

NEWS
```diff
@@ -6,6 +6,15 @@ Please send bug reports, questions and suggestions to
...
+  The encrypt{,_r} function will - for security reasons - additionally
+  overwrite the data-block argument with random data.
```
zackw: "For security reasons, the encrypt{,r} functions will also overwrite their data-block argument with random bits."

NEWS
```diff
@@ -6,6 +6,15 @@ Please send bug reports, questions and suggestions to
...
+  In case of the fcrypt function a unique failure-token (*0 or *1)
+  will be returned; when failure-tokens have been disabled during
+  build-time fcrypt will always return NULL.
```
zackw: "The fcrypt function will always produce a failure token (*0 or *1), unless the library was also configured with --disable-failure-tokens, in which case it will always return NULL."

configure.ac
```diff
@@ -304,6 +304,23 @@ AC_ARG_ENABLE([obsolete-api],
...
 AC_SUBST(COMPAT_ABI)
 
+AC_ARG_ENABLE([obsolete-api-enosys],
+    AS_HELP_STRING(
+        [--enable-obsolete-api-enosys[=ARG]],
+        [If set to "yes", this setting replaces the obsolete APIs (fcrypt,
+         encrypt{,_r}, and setkey{,_r}) with POSIX compliant stubs setting
+         "errno" to "ENOSYS", instead of actually performing any real
+         operation.  The encrypt{,_r} function will - for security reasons
+         - additionally overwrite the data-block argument with random data.
```
zackw: Same change suggested here as in NEWS.

configure.ac
```diff
@@ -331,21 +348,41 @@ if test x"$hashes_enabled" = x || test x"$hashes_enabled" = x,; then
 fi
 AC_SUBST([hashes_enabled])
 
+
 # If the traditional DES hash is disabled, then the obsolete APIs are
-# implicitly disabled.
+# implicitly disabled, except for the case stubs are requested.
```
zackw: ", except when stubs are requested."

configure.ac
```diff
@@ -331,21 +348,41 @@ if test x"$hashes_enabled" = x || test x"$hashes_enabled" = x,; then
...
     *,descrypt,*) ;;
     *)
        if test x"$COMPAT_ABI" != xno; then
-         AC_MSG_WARN(
-           [--enable-hashes=${hashes_selected} forces --enable-obsolete-api=no])
+         if test x"$enable_obsolete_api_enosys" != x0; then
+           AC_MSG_WARN(
+             [--enable-hashes=${hashes_selected} does NOT include "descrypt".
+              The crypt{,_r{,a,n}} functions will not be POSIX compliant, nor
+              fully backwards compatible.])
```
zackw: POSIX does not require `crypt*` to support descrypt-based passphrase hashing (see http://pubs.opengroup.org/onlinepubs/9699919799/functions/crypt.html -- an implementation that always returns ENOSYS when the salt argument is in the specified form is valid).  I don't think this warning is necessary; someone who turns off DES knows what they're doing.

crypt-common.h
```diff
@@ -30,4 +30,24 @@ const unsigned char ascii64[65];
 #define b64t   ((const char *) ascii64)
 #define itoa64 ascii64
 
+/* For historical reasons, crypt and crypt_r are not expected ever to
+   return 0, and for internal implementation reasons (see
+   call_crypt_fn, below), it is simpler if the individual algorithms'
```
zackw: Cut-and-paste error here: the parenthetical should read "see call_crypt_fn, in crypt.c".

crypt-des-obsolete.c
```diff
@@ -116,17 +124,26 @@ do_setkey_r (const char *key, struct des_ctx *ctx)
...
 
 #if INCLUDE_setkey_r
+#if ENABLE_OBSOLETE_API_ENOSYS
+void
+setkey_r (ARG_UNUSED (const char *key), ARG_UNUSED (struct crypt_data *data))
+{
+  /* Pretend this function is not supported.  */
+  errno = ENOSYS;
+}
+#else
```
zackw: It's fine to put `ARG_UNUSED` on an argument that _may_ be used, and I don't like duplicating an entire function definition in two branches of a preprocessor conditional, I worry that the prototypes will get out of sync.  Also, in this configuration, we're not _pretending_ the function isn't supported.  Suggest instead

``` c
#if INCLUDE_setkey_r
void
setkey_r (ARG_UNUSED (const char *key), ARG_UNUSED (struct crypt_data *data))
{
#if ENABLE_OBSOLETE_API_ENOSYS
  /* This function is not supported in this configuration.  */
  errno = ENOSYS;
#else
  do_setkey_r (key, get_des_ctx (data));
#endif
}
SYMVER_setkey_r;
#endif
```

crypt-des-obsolete.c
```diff
@@ -138,11 +155,28 @@ do_encrypt_r (char *block, int edflag, struct des_ctx *ctx)
...
+     case get_random_bytes() fails.  */
+  XCRYPT_SECURE_MEMSET(block, 64);
+
+  /* Overwrite sensitive data with random data.  */
+  get_random_bytes(block, 64);
+
+  /* Pretend this function is not supported.  */
+  errno = ENOSYS;
+}
+#else
```
zackw: Same concern here as for `setkey_r`.

crypt-des-obsolete.c
```diff
@@ -152,24 +186,49 @@ SYMVER_encrypt_r;
...
 
 #if INCLUDE_setkey
+#if ENABLE_OBSOLETE_API_ENOSYS
+void
+setkey (ARG_UNUSED (const char *key))
+{
+  /* Pretend this function is not supported.  */
+  errno = ENOSYS;
+}
+#else
```
zackw: Same concern here as for `setkey_r`.

crypt-des-obsolete.c
```diff
@@ -152,24 +186,49 @@ SYMVER_encrypt_r;
...
+     case get_random_bytes() fails.  */
+  XCRYPT_SECURE_MEMSET(block, 64);
+
+  /* Overwrite sensitive data with random data.  */
+  get_random_bytes(block, 64);
+
+  /* Pretend this function is not supported.  */
+  errno = ENOSYS;
+}
+#else
```
zackw: Same concern here as for `setkey_r`.

crypt-des-obsolete.c
```diff
@@ -152,24 +186,49 @@ SYMVER_encrypt_r;
...
 SYMVER_setkey;
 #endif
 
 #if INCLUDE_encrypt
+#if ENABLE_OBSOLETE_API_ENOSYS
+void
+encrypt (char *block, ARG_UNUSED (int edflag))
+{
+  /* Make sure sensitive data is at least zero'ed in
+     case get_random_bytes() fails.  */
```
zackw: Grammar nit: in formal register `zero'ed` is incorrect.  I would write this comment "Make sure sensitive data is erased in case get_random_bytes fails."

crypt-static.c
```diff
@@ -17,6 +18,13 @@
 #include "crypt-port.h"
 #include "xcrypt.h"
 
+#if ENABLE_OBSOLETE_API_ENOSYS
+char *
+fcrypt (const char *key, const char *setting);
```
zackw: Why is it necessary to add a prototype for fcrypt?  Also, putting the function name on the line after the return type is only for function _definitions_ (the point of that convention is you can do `grep '^function_name' *.c` to find only the definition).  Also, the inclusion of errno.h should come first.

crypt-static.c
```diff
@@ -35,7 +43,30 @@ SYMVER_crypt;
...
+#if ENABLE_FAILURE_TOKENS
+  /* Return static buffer filled with a failure-token.  */
+  static char retval[3];
+  make_failure_token (setting, retval, 3);
+  return retval;
+#else
+  return NULL;
+#endif
+}
+#else
```
zackw: Same concern here as for `setkey_r`.

crypt-des-obsolete.c
```diff
@@ -49,7 +52,12 @@
 #include "crypt-obsolete.h"
 #include "alg-des.h"
 
-#if INCLUDE_encrypt || INCLUDE_encrypt_r || INCLUDE_setkey || INCLUDE_setkey_r
+#if ENABLE_OBSOLETE_API_ENOSYS
+#include <errno.h>
+#endif
```
zackw: Throughout: it seems like it would be more readable and less of a future trip hazard if we just include errno.h unconditionally in each file that might need it.

test-fcrypt-enosys.c
```diff
@@ -0,0 +1,82 @@
...
+  return status;
+}
+
+#else
+
+int
+main (void)
+{
+  return 77; /* UNSUPPORTED */
+}
```
zackw: Hmm, when `ENABLE_OBSOLETE_API_ENOSYS` is _not_ true, should this test that `fcrypt` does the same thing as `crypt`?  I don't believe we have any other tests of `fcrypt`.

solardiz:
This sounds dangerous to me, to the point where I think I'd NAK this if I were the maintainer. Programs that actually use these APIs will fail in dangerous ways, such as not encrypting data where they previously would have, or e.g. a VNC server not performing proper authentication.

crypt-common.h
besser82: Fixed.

crypt-static.c
besser82: > Why is it necessary to add a prototype for fcrypt?

We compile with `-Wmissing-prototypes -Werror` by default.  For that reason we need to prototype `fcrypt` as soon as we don't declare it a `strong_alias` for `crypt`.


> Also, putting the function name on the line after the return type is only for function _definitions_ (the point of that convention is you can do `grep '^function_name' *.c` to find only the definition). Also, the inclusion of errno.h should come first.

Fixed.

besser82: I've moved the prototype for `fcrypt` with a conditional into `crypt-port.h`.

crypt-des-obsolete.c
besser82: Fixed.

crypt-des-obsolete.c
besser82: Fixed.

crypt-des-obsolete.c
besser82: Fixed.

crypt-des-obsolete.c
besser82: Fixed.

crypt-des-obsolete.c
besser82: Fixed.

crypt-static.c
besser82: Why?  We only define `fcrypt` as function, when we build it as a stub.  Otherwise it is a `strong_alias` for `crypt`.

configure.ac

configure.ac

Shall `crypt*` return `ENOSYS` when trying to use `descrypt`, when stubs are built and `descrypt` is disabled?

NEWS
besser82: Fixed.

NEWS
besser82: Fixed.

NEWS
besser82: Fixed.

configure.ac

test-fcrypt-enosys.c
besser82: I think we should add tests for that case, but I'll do that in second PR / separate commit.

besser82:
@zackw I've pushed an updated / rebased commit, which addresses all you suggestions, except the additional testcase for `fcrypt`.  I'll add that in a separate future PR.

besser82:
@solardiz I've updated the tests to explicitly reset `errno = 0` in the rebased commit.

zackw:
(Accidentally closed this by clicking in the wrong place, sorry about that.)

Github’s web interface is malfunctioning for me, so I am responding to a bunch of stuff all at once instead of in line.

@besser82

> > Why is it necessary to add a prototype for fcrypt?
> We compile with `-Wmissing-prototypes -Werror` by default. For that reason we need to prototype fcrypt as soon as we don't declare it a strong_alias for crypt.

Oh right, I forgot about `-Wmissing-prototypes`.

> I've moved the prototype for fcrypt with a conditional into crypt-port.h.

That sounds good.

> We only define fcrypt as function, when we build it as a stub. Otherwise it is a strong_alias for crypt.

Yes, but you still have two different versions of the function head that only differ in usage of `ARG_UNUSED`, that’s what I’d prefer not to have.

> Shall `crypt*` return ENOSYS when trying to use descrypt, when stubs are built and descrypt is disabled?

I think it is less likely to cause problems if we stick to EINVAL for this case.

> I think we should add tests for [fcrypt being the same as crypt when not `--enable-obsolete-api-enosys`], but I'll do that in second PR / separate commit.

Ok.

> I've updated the tests to explicitly reset errno = 0 in the rebased commit.

Thanks, I agree that’s important.

@solardiz

> Programs that actually use these APIs will fail in dangerous ways, such as not encrypting data where they previously would have

I am a little worried about a hypothetical program that uses `encrypt()` for encryption of data to be stored, calls it with no error checking, and destroys the cleartext.  Before this change, the data was encrypted insecurely.  After, it’s destroyed permanently, which may be worse.

I went through https://codesearch.debian.net/ looking for uses of encrypt and/or setkey and I did find such a program: [cairo-dock](https://sources.debian.org/src/cairo-dock/3.4.1-3/src/gldit/cairo-dock-config.c/?hl=526#L487) appears to be using the DES encryption functions to store passwords for email accounts on disk.  (I have reported this as a bug in cairo-dock.)

Given that any use of either encrypt or setkey in a production program nowadays is in and of itself a bug, though, I think we can live with the risk.

crypt-static.c
besser82: I can just find one function head for `fcrypt` here…  :/

zackw: As of 0919e33d0e12bf580ce389456e73ce3b88afd771 it seems to be fine.

solardiz:
@zackw I'm not sure use of these functions is necessarily a bug. There are still network protocols, etc. that use DES - a notable example is VNC, which uses DES in its challenge-response authentication. Arguably, such programs should include their own implementation of DES these days.

besser82:
Allrighty.  Waiting for Travis to pass.  Will merge on CLI then.

besser82:
@solardiz FYI, I'm going to make use of this configure option in Fedora 30:  https://pagure.io/fesco/issue/2062

zackw:
@solardiz Yeah, IMHO any program that still needs to use DES for whatever reason should include its own implementation (or use e.g. OpenSSL's implementation).  You already can't link new programs against them if the system libcrypt is libxcrypt.

PR #79, opened 2019-02-01
Testsuite consolidation

zackw:
Most of the hashing methods had their own test-crypt-xxx.c program that ran a short sequence of “known answer” tests: `crypt(phrase, salt) = expected` for fixed values of phrase, salt, and expected.  The code involved was very repetitive, taken as a whole, and many of the programs were not very thorough.

This patchset consolidates all of those programs into a single one that is much more thorough.  It may actually be _too_ thorough - it takes 20 seconds to run, on my computer, and that's after I parallelized it.  Under `valgrind` it takes upward of 20 *minutes* to run, which necessitated some hacking around Travis's default timeouts.  We could speed it up by dropping the checks that `crypt`, `crypt_r`, `crypt_rn`, and `crypt_ra` all produce the same output for the same input.  If we look inside the black box, we know from the structure of the code that this will always be true for valid inputs (because the other three are wrappers around `crypt_rn`).  I am arguing with myself about whether it is safe to know that for testing purposes, and would appreciate a second opinion.

Please also check my logic regarding the somewhat weaker treatment of (gost-)yescrypt than the other hashes (see the individual commit comments, and the notes about yescrypt in `test-crypt-kat-gen.py`).


besser82:
Added `test-crypt-kat` to `.gitignore`.
Reverted the truncation for NTHASH.

zackw:
Thanks. I have a plan for what to do about yescrypt but it’s going to take
me a couple more days to execute.

>

zackw:
@besser82
OK, I'm now generating (gost-)yescrypt hashes using ctypes to call into our libcrypt.so, which will, as you suggest, verify hash stability - the behavior now is the same as the behavior as of the last time `test-crypt-kat.inc` was regenerated (and git will tell us if a change to the generator has unexpectedly changed the expected hashes).

Before merging, I want to ask again whether you think it is worthwhile to test that all four of `crypt`, `crypt_r`, `crypt_rn`, and `crypt_ra` produce the same output for the same input for all of the test patterns.  This is the major reason test-crypt-kat is so slow and there's an argument that it's pointless since we know they all just call `do_crypt`.

besser82:
> @besser82
> OK, I'm now generating (gost-)yescrypt hashes using ctypes to call into our libcrypt.so, which will, as you suggest, verify hash stability - the behavior now is the same as the behavior as of the last time `test-crypt-kat.inc` was regenerated (and git will tell us if a change to the generator has unexpectedly changed the expected hashes).

Sounds good!


> Before merging, I want to ask again whether you think it is worthwhile to test that all four of `crypt`, `crypt_r`, `crypt_rn`, and `crypt_ra` produce the same output for the same input for all of the test patterns. This is the major reason test-crypt-kat is so slow and there's an argument that it's pointless since we know they all just call `do_crypt`.

I think it is a good idea to check whether all of the crypt functions are acting the same.  For completeness, we should include a consistency test for `fcrypt` (if built as an alias for `crypt`) as well as you suggested in my PR about the optional replacement stubs.

PR #81, opened 2019-02-04
Add cosistency test for fcrypt.

besser82:
As suggested / proposed in #79.


besser82:
Looks like we need to get #62 working before we can merge this one, as we cannot run `fcrypt` in a parallel thread aside of `crypt`.  Running them sequentially makes the valgrind test with Clang timeout.  :/

besser82:
@zackw I've corrected the way `CFLAGS` are set.

besser82:
@zackw Added `travis_wait 60` to `make check`.

besser82:
Will merge on CLI as soon as Travis turns green.

PR #82, opened 2019-02-06
Some improvements to Autotools.

besser82:
Basically this adds some sanity checks in configure around the `regen-test-crypt-kat` target in Makefile.


besser82:
@zackw Are you fine with this change?

besser82:
Rebased onto develop, squashed all previous commits into one.

configure.ac
```diff
@@ -238,6 +238,39 @@ AX_VALGRIND_DFLT([sgcheck], [off])
 # Add a target to run testsuite with valgrind.
 AX_VALGRIND_CHECK()
 
+# Requirements for running test-crypt-kat-gen.py.
+ENABLE_CRYPT_KAT_GEN="no"
+PYTHON_BIN="python3 python"
+
+for ver in `seq 0 1 10`; do
+  PYTHON_BIN="python3.$ver $PYTHON_BIN"
```
zackw: Shouldn't this be `seq 6 1 10` since we already know 3.5 and lower won't work?

Also I suggest `PYTHON_BIN` should be called `PYTHON_CANDIDATES` instead.  `PYTHON_BIN` feels too likely to collide with some build robot's environment variables.

configure.ac
```diff
@@ -238,6 +238,39 @@ AX_VALGRIND_DFLT([sgcheck], [off])
...
+
+AC_PATH_PROGS([PYTHON], [$PYTHON_BIN], [])
+
+if test x"$PYTHON" != "x"; then
+  AX_PROG_PYTHON_VERSION([3.6], [ENABLE_CRYPT_KAT_GEN="yes"],
+    [AC_MSG_WARN([Python >= 3.6 is required for optional Makefile targets.])
+  ])
+
+  if test x"$ENABLE_CRYPT_KAT_GEN" = "xyes"; then
+    for pymod in array ctypes hashlib multiprocessing os passlib re sys; do
```
zackw: It should only be necessary to do this test for `passlib`.  All the other modules are part of the Python standard library.

configure.ac
```diff
@@ -238,6 +238,39 @@ AX_VALGRIND_DFLT([sgcheck], [off])
...
+  PYTHON_BIN="python3.$ver $PYTHON_BIN"
+done
+
+AC_PATH_PROGS([PYTHON], [$PYTHON_BIN], [])
+
+if test x"$PYTHON" != "x"; then
+  AX_PROG_PYTHON_VERSION([3.6], [ENABLE_CRYPT_KAT_GEN="yes"],
+    [AC_MSG_WARN([Python >= 3.6 is required for optional Makefile targets.])
+  ])
+
```
zackw: This works because you set up the order of candidates in `PYTHON_BIN` to go from most to least likely to be the one we want.  That probably deserves a comment, so future maintainers don't have to work out why it's not necessary to loop over the contents of `PYTHON_BIN` calling `AC_PATH_PROG` and then `AX_PROG_PYTHON_VERSION` for each.

configure.ac

configure.ac

besser82:
Fixes are in the rebased commit.  Will merge on CLI as soon as Travis turns green.

configure.ac
besser82: Fixed by adding a comment above.

PR #84, opened 2019-03-01
Move code, documentation, and tests into subdirectories.

zackw:
There are enough files now that it’s getting confusing to have
everything at the top level.  Create subdirectories ‘lib’, ‘doc’, and
‘test’.  Move all of the code linked into libcrypt.a into ‘lib’. Move
all the manpages into ‘doc’.  Move all of the test programs into ‘test’.

There is still only one Makefile at top level.  Automake doesn’t make
nonrecursive makefiles as easy as it could, but everything that was
written in http://aegis.sourceforge.net/auug97.pdf back in 1997(!) is
still true.
https://www.microsoft.com/en-us/research/wp-content/uploads/2016/03/hadrian.pdf
has an interesting counterpoint but I don’t think we’re anywhere near
the scale where those problems are relevant.


zackw:
Thanks, and sorry for forgetting about .gitignore.

PR #86, opened 2019-05-03
gost-yescrypt: Fix alignment problem for GOST 34.11 (Streebog)

vt-alt:
Some architectures in some circumstances do not allow unaligned
memory access (such as ARM, MIPS, SPARC) triggering SIGBUS. This
patch very crudely fixes this issue.

The issue is found and original fix is proposed by Eric Biggers:

  https://patchwork.kernel.org/patch/10878865/

Being unfixed this would trigger SIGBUS when password buffer is
unaligned. Crash and fix are tested on UltraSparc T5 on GCC Compile
farm.

### DIFF
--- a/lib/alg-gost3411-2012-core.c
+++ b/lib/alg-gost3411-2012-core.c
@@ -149,10 +149,13 @@ g(uint512_u *h, const uint512_u *N, const unsigned char *m)
 static inline void
 stage2(GOST34112012Context *CTX, const unsigned char *data)
 {
-    g(&(CTX->h), &(CTX->N), data);
+    union uint512_u m;
+
+    memcpy(&m, data, sizeof(m));
+    g(&(CTX->h), &(CTX->N), (const unsigned char *)&m);
 
     add512(&(CTX->N), &buffer512, &(CTX->N));
-    add512(&(CTX->Sigma), (const uint512_u *) data, &(CTX->Sigma));
+    add512(&(CTX->Sigma), &m, &(CTX->Sigma));
 }
 
 static inline void


zackw:
Does the existing test suite detect this problem when run on hardware that requires alignment of this buffer?  If not, please add a test case.

vt-alt:
@zackw Do you want a general test in `crypt-kat.c` or just for `crypt-gost-yescrypt.c`? I was quick tested with this:
```
vt@gcc202:~/src/libxcrypt$ git diff
diff --git test/crypt-kat.c test/crypt-kat.c
index 0cd8c2b..253b0a8 100644
--- test/crypt-kat.c
+++ test/crypt-kat.c
@@ -207,13 +207,17 @@ calc_hashes_crypt_r_rn (ARG_UNUSED (void *unused))
   memset (&data, 0, sizeof data);
   for (i = 0; i < ntests; i++)
     {
+       char pass[CRYPT_MAX_PASSPHRASE_SIZE];
+
+       strcpy(pass + 1, tests[i].input);
+       printf("[%d]: %s %s\n", strlen(tests[i].input), tests[i].input, tests[i].salt);
       errno = 0;
-      hash = crypt_r (tests[i].input, tests[i].salt, &data);
+      hash = crypt_r (pass + 1, tests[i].salt, &data);
       status |= report_result ("crypt_r", hash, errno, &tests[i],
                                ENABLE_FAILURE_TOKENS);

       errno = 0;
-      hash = crypt_rn (tests[i].input, tests[i].salt, &data, (int)sizeof data);
+      hash = crypt_rn (pass + 1, tests[i].salt, &data, (int)sizeof data);
       status |= report_result ("crypt_rn", hash, errno, &tests[i], false);
     }
```
Only `gost-yescrypt` triggered an error.

Currently, `calc_hashes_crypt_r_rn` does not output anything, silent test, so when it crashes it's unknown why - that's why there is `printf`. Also, you don't have an error message, just a crash with `Bus error`.

vt-alt:
> Does the existing test suite detect this problem when run on hardware that requires alignment of this buffer?

Forgot to say 'no'.

besser82:
@zackw @vt-alt LGTM.  

> @zackw Do you want a general test in `crypt-kat.c` or just for `crypt-gost-yescrypt.c`? I was quick tested with this:
> 
> ```
> vt@gcc202:~/src/libxcrypt$ git diff
> diff --git test/crypt-kat.c test/crypt-kat.c
> index 0cd8c2b..253b0a8 100644
> --- test/crypt-kat.c
> +++ test/crypt-kat.c
> @@ -207,13 +207,17 @@ calc_hashes_crypt_r_rn (ARG_UNUSED (void *unused))
>    memset (&data, 0, sizeof data);
>    for (i = 0; i < ntests; i++)
>      {
> +       char pass[CRYPT_MAX_PASSPHRASE_SIZE];
> +
> +       strcpy(pass + 1, tests[i].input);
> +       printf("[%d]: %s %s\n", strlen(tests[i].input), tests[i].input, tests[i].salt);
>        errno = 0;
> -      hash = crypt_r (tests[i].input, tests[i].salt, &data);
> +      hash = crypt_r (pass + 1, tests[i].salt, &data);
>        status |= report_result ("crypt_r", hash, errno, &tests[i],
>                                 ENABLE_FAILURE_TOKENS);
> 
>        errno = 0;
> -      hash = crypt_rn (tests[i].input, tests[i].salt, &data, (int)sizeof data);
> +      hash = crypt_rn (pass + 1, tests[i].salt, &data, (int)sizeof data);
>        status |= report_result ("crypt_rn", hash, errno, &tests[i], false);
>      }
> ```
> 
> Only `gost-yescrypt` triggered an error.
> 
> Currently, `calc_hashes_crypt_r_rn` does not output anything, silent test, so when it crashes it's unknown why - that's why there is `printf`. Also, you don't have an error message, just a crash with `Bus error`.

The test-case seems fine to me this way as well.  Maybe it can be optimized a bit, so that the buffer won't be allocated in every iteration of the loop.  Also we may want to use `strncpy(pass + 1, tests[i].input, CRYPT_MAX_PASSPHRASE_SIZE - 1);`, so the buffer is guaranteed to be filled with trailing zeros.

Anyways, I'm going to modify the test-case myself after merging fix.

vt-alt:
@besser82 
> Anyways, I'm going to modify the test-case myself after merging fix.
Ah, I will at least test it.

zackw:
I do want each hash algorithm to be tested at least once for this bug.  I doubt we get any more test coverage out of feeding them misaligned input buffers for every test in test-crypt-kat.c, but if that's the easiest way to accomplish the test then I don't see any harm in it.

lib/alg-gost3411-2012-core.c
```diff
@@ -149,10 +149,13 @@ g(uint512_u *h, const uint512_u *N, const unsigned char *m)
 static inline void
 stage2(GOST34112012Context *CTX, const unsigned char *data)
 {
-    g(&(CTX->h), &(CTX->N), data);
+    union uint512_u m;
```
solardiz: Looks like the word `union` here is unneeded (implied as part of the `typedef`), right? If so, I'm mildly surprised compilers are happy to ignore it.

besser82: Fixed in 5cce1ebe4e8e303bab593af694c9d30c51cf5eb4.

solardiz:
BTW, `pass + 1` doesn't _guarantee_ misalignment, it just makes misalignment very likely. We could want to test e.g. with both `pass` and `pass + 1`, so that one of those is certainly misaligned, or test with the right one of them after explicitly testing for misalignment (cast to `uintptr_t` and test the least significant bit).

solardiz:
Something like:

```c
mispass = pass;
if (!((uintptr_t)mispass & 1))
    mispass++;
```

or:

```c
mispass = pass + 1;
if (((uintptr_t)pass & 1))
    mispass = pass;
```

then use `mispass` in place of the uses of `pass + 1` proposed above.

vt-alt:
> BTW, pass + 1 doesn't guarantee misalignment

Oh ya since `pass` is `char` buffer it does not need to be aligned. Maybe we could just union it to `int`?

solardiz:
> Maybe we could just union it to `int`?

That's also a valid fix, yes.

besser82:
> > BTW, pass + 1 doesn't guarantee misalignment
> 
> Oh ya since `pass` is `char` buffer it does not need to be aligned. Maybe we could just union it to `int`?

That does not work as it will trigger `-Wstringop-overflow=` when using `strcpy`…

vt-alt:
@besser82 I pushed in #87 my version of it.

vt-alt:
I tested on UltraSparc T5 again.
- Current `develop` branch produces no errors.
- Then I reverted commits 5cce1eb and 03dccc3 with Streebog fixes and got
```
vt@gcc202:~/src/libxcrypt$ test/crypt-kat
...
[80]: 12345678901234567890123456789012345678901234567890123456789012345678901234567890 $gy$j75$.......
Bus error
vt@gcc202:~/src/libxcrypt$ make check
...
PASS: test/crypt-gost-yescrypt
./m4/test-driver: line 107: 195553 Bus error               "$@" > $log_file 2>&1
FAIL: test/crypt-kat
...
```
Thus we know that `test/crypt-kat` is correct.

PR #92, opened 2019-09-01
Add binary compatibility for GNU/Hurd and GNU/kFreeBSD

### LINKED ISSUE #91: Please enable binary compatibility for GNU/Hurd as well
Hello,

I see that libxcrypt assumes GNU/Hurd doesn't use binary compatibility

https://github.com/besser82/libxcrypt/blob/3add78ffac30ca3785eea5cd35619710b50432b0/configure.ac#L464

but it does :) (at least formally with abilists since glibc 2.27). We have thousands of Debian binary packages which we want to see continue working when Debian's glibc will migrate to libxcrypt.

The crypt interface on GNU/Hurd is basically the same as on GNU/Linux, since it's the same source code, so everything should be "just working" by using the same codepaths.

The same holds true for GNU/kFreeBSD as well.

I'll submit a MR to fix this.

Samuel

sthibaul:
Fixes #91


besser82:
LGTM!  Let's wait for TravisCI to go green, and I'll merge.  After that I'll release a v4.4.8 for your convenience.

sthibaul:
Thanks!

Issue #96, opened 2019-12-28
Wrong SONAME for alpha and ia64


rfc1036:
A Debian porter noticed that alpha and ia64 use `libcrypt.so.1.1` instead of `libcrypt.so.1`.
I am not even sure that it is possible for libcrypt to use a SONAME like this, so for the time being I am patching libtool in the build directory:

`perl -i -pe '/^(library_names_spec|soname_spec)=/ and s/major/major.1/'`

zackw:
Good catch, thanks.  Unfortunately I don't think libtool supports sonames like this at all.  I've got an experimental branch in which we build with meson instead, and I'll see if that can do it.  (However, is it a potential problem for porters and/or new architectures if we use meson instead of autotools?  You'd know better than me/)

rfc1036:
I have no experience at all with Meson, but since it is written in pure Python I suppose that it works on all Debian architectures.

besser82:
From the perspective of bootstrapping new architectures from scratch there shouldn't be any problems as they need to be cross-compiled, which is perfectly supported by meson, on some existing system with a working Python3 to get a basic stage-1 image/tarball anyways.

rfc1036:
Actually the soname is wrong only on alpha, ia64 is "standard".

mattst88:
I can confirm that glibc installs `libcrypt.so.1.1` on alpha (and libc.so.6.1, presumably due to the same SONAME bump long ago):

````sh
$ qlist glibc | grep libcrypt.so.1
/lib/libcrypt.so.1.1
/usr/lib/libcrypt.so
````

while libxcrypt installs `libcrypt.so.1`:

````sh
$ qlist libxcrypt | grep libcrypt.so.1
/lib/libcrypt.so.1.1.0
/lib/libcrypt.so.1
````

On Gentoo, we rebuilt reverse dependencies of libcrypt when doing this transition, so it was not a problem for us. I believe this would have been a problem for a binary distribution that would necessitate a rebuild, since the SONAME in binaries is indeed `libcrypt.so.1.1`:

````sh
$ readelf -a /usr/sbin/sshd  |& grep libcrypt.so
 0x0000000000000001 (NEEDED)             Shared library: [libcrypt.so.1.1]
  000000: Version: 1  File: libcrypt.so.1.1  Cnt: 1
````
On ia64, glibc installs `libcrypt.so.1` and libxcrypt does the same:

````sh
$ qlist glibc | grep libcrypt.so.1
[...]

PR #97, opened 2020-01-17
Convert build system from autotools to meson.

zackw:
Meson (https://mesonbuild.com/) is a relatively new configuration and build utility for C programs.  It offers several benefits for us: Its scripting language is substantially nicer to work with than either autoconf’s shell+m4 or makefiles.  It seems to be faster, particularly for the configure phase.  Its shared library build logic, unlike libtool, will let us fix issue 96 properly [note: this PR does _not_ fix issue 96].  And, perhaps most importantly, it seems to be more actively maintained than autoconf, which we were lugging around nearly 10,000 lines worth of patches and extensions for.

Meson does have some disadvantages: most importantly, it depends on Python >=3.6, and a tool called ‘ninja’ (https://ninja-build.org/) for the actual building.  Because it’s under active development (unlike autotools) we have to worry about having a suitably new version of meson itself installed.  [I have not actually tested this PR with anything older than 0.53, but I believe the newest features we use come from 0.50.]  It doesn’t have a ‘distcheck’ equivalent, and it has a number of annoying bugs that I had to work around (see comments in the meson.build files).  All of our packagers will have to adapt their build machinery.

Because this patch introduces a build-time requirement for Python >=3.6 *anyway*, I also rewrote all of the build scripts in that language, which eliminates the dependency on AWK and comes out a whole lot nicer to read, too.  The only exceptions are scripts/travis-install, which runs before we have a known-good Python environment, and scripts/travis-after-success, which is so simple that it’s easier to keep it in sh.

All of our existing build configuration options are preserved, but they might be spelled differently now.  See meson_options.txt at top level.  .travis.yml and scripts/travis-build may also be useful to crib from.

This does not by itself fix the failing MacOS build, but the remaining problems are portability bugs in the actual code rather than bugs in the build scripts, so they should be addressed separately.

----
I've been working on this for a while and I think it's now ready for review. Besides standard code review and weighing in on the obvious question of whether we should make this change at all, I'd like to ask for testing in unusual configurations (if it's not covered by the Travis CI builds, you can assume I haven't tried it either).

NOTE: this patch includes a NEWS entry for version "4.5.0" but I have plans for several follow-up patches after this one and before releasing that version.  I'll write up those plans for your consideration this weekend.


besser82:
> I've been working on this for a while and I think it's now ready for review. Besides standard code review and weighing in on the obvious question of whether we should make this change at all, I'd like to ask for testing in unusual configurations (if it's not covered by the Travis CI builds, you can assume I haven't tried it either).

Quick checks on possibly untested scenarios:

* Building with `meson … --unity on` doesn't work, yet.  (I have a patch-set ready, tho.)
* The installed shared libs are named differently now: `libcrypt.so.1` vs. `libcrypt.so.1.1.0` and `libcrypt.so.2` vs. `libcrypt.so.2.0.0`.  (Can be fixed easily, see: https://mesonbuild.com/Reference-manual.html#shared_library.  I'll take care of that, too.)


Other things that need love:

* The rpm spec file template needs updating.  (Will take care of that before merging.)
* Some logic in the meson build files might be simplified.


Anyways, I'm +1 for changing our buildsystem to Meson.

besser82:
Another finding:  Codecov data isn't uploaded for all configurations anymore.

zackw:
> Building with `meson … --unity on` doesn't work, yet.  (I have a patch-set ready, tho.)

Yeah, I saw you were working on that.  I don’t think we need that to work for the initial merge?  I wrote in the README that it’s known not to work.

> The installed shared libs are named differently now: `libcrypt.so.1` vs. `libcrypt.so.1.1.0` and `libcrypt.so.2` vs. `libcrypt.so.2.0.0`.  (Can be fixed easily, see: https://mesonbuild.com/Reference-manual.html#shared_library.  I'll take care of that, too.)

Just a matter of setting `version: '1.1.0'` or `version: '2.0.0'` right?  I can add that to the next revision of the patch.  I wonder how much it matters, though.  Nothing should refer to the `libcrypt.so.X.Y.Z` names directly.

> The rpm spec file template needs updating.  (Will take care of that before merging.)

I’d appreciate it if you could handle that part, I don’t know anything about RPM.

> Some logic in the meson build files might be simplified.

Could you be more specific?  I was learning the meson language as I went along, so for sure there might be improvements to be made.

> Codecov data isn't uploaded for all configurations anymore.

Oh, is *that* why coverage dropped by 11%?  I thought we only needed to upload coverage for the `-O0` builds, and the change report didn’t make a whole lot of sense.  I’ll tinker with this.

besser82:
> > Building with `meson … --unity on` doesn't work, yet.  (I have a patch-set ready, tho.)
> 
> Yeah, I saw you were working on that. I don’t think we need that to work for the initial merge? I wrote in the README that it’s known not to work.

No, that would be seperate PR.


> > The installed shared libs are named differently now: `libcrypt.so.1` vs. `libcrypt.so.1.1.0` and `libcrypt.so.2` vs. `libcrypt.so.2.0.0`.  (Can be fixed easily, see: https://mesonbuild.com/Reference-manual.html#shared_library.  I'll take care of that, too.)
> 
> Just a matter of setting `version: '1.1.0'` or `version: '2.0.0'` right? I can add that to the next revision of the patch. I wonder how much it matters, though. Nothing should refer to the `libcrypt.so.X.Y.Z` names directly.

Well, I investigated a bit more about that, and found out, glibc only installs `libcrypt.so.1` (except for some niche arches, where it is `libcrypt.so.1.1`).  So I guess, we don't need any deeper versioning either, tho we shipped `so.1.1.0` and `so.2.0.0` with Autotools based builds.

> > The rpm spec file template needs updating.  (Will take care of that before merging.)
> 
> I’d appreciate it if you could handle that part, I don’t know anything about RPM.

Done so.  I will get that reviewed within Fedora and make suggested adjustedments in seperate commits outside of this PR.  The spec file builds with the changes I made, and its sole purpose is to do test builds in a Fedora COPR (think of it to be the equivalent to a PPA in Ubuntu universe).

> > Some logic in the meson build files might be simplified.
> 
> Could you be more specific? I was learning the meson language as I went along, so for sure there might be improvements to be made.

Meson >= 0.50 ships build-in modules for detecting the version of the Python interpreter and compare versions.  But things like that can be subject to later commits outside of this PR.

> > Codecov data isn't uploaded for all configurations anymore.
> 
> Oh, is _that_ why coverage dropped by 11%? I thought we only needed to upload coverage for the `-O0` builds, and the change report didn’t make a whole lot of sense. I’ll tinker with this.

Yes, previously I submitted coverage data for each build on TravisCI, except COVERITY, Valgrind and ASAN/UBSAN configuration.  That is the reason why coverage dropped now.

besser82:
> Couple of notes specifically on [357ea1a](https://github.com/besser82/libxcrypt/commit/357ea1a6a35d78536814f26a35fdc36ab2fb0e2e):
> 
> 1. Can you please explain why you want a trailing slash on the name of the build directory, whereever it's used?  It doesn't seem to be necessary.

I done have any specific reason, but explicitly showing we are operating on a directory here.  Besides this I want to clarify `build` is actually not an internal meson command, which might be ambiguous otherwise.

> 2. If we're going to name both the source and the build directory in the `meson setup` phase, which makes sense for explicitness, perhaps we should also specifically invoke `meson setup` instead of just `meson`?

I do not have an opinion on this, as meson use that syntax in its documentation, but does _not_ use it in its [rpm macros](https://github.com/mesonbuild/meson/blob/master/data/macros.meson).

> I'm not qualified to review [9afe3c8](https://github.com/besser82/libxcrypt/commit/9afe3c8c1ed5580e7c02c60f18d6a8270e243679).

Will be done by fellow Fedora contributors.  See my other reply above.

***

BTW, I just discovered the Autotools `dist-check` equivalent for Meson:

```
meson . build/
meson dist -C build/
```

Which will build, install and test the project and drop a xz-compressed source-tarball (with a SHA256 checksum file) in `build/meson-dist`.

besser82:
* Updated project version in `meson.build`:  4.4.11 -> 4.5.0
* Explicitly use `meson setup` for the configure stage.

zackw:
I just noticed that there's a bug in the generation of libcrypt.map with `-Dobsolete-api=false`.  It doesn't leave out all of the compatibility symbols in that case.  I'll look into this.

besser82:
* Increased timeout-multiplier in rpkg spec file template for Valgrind: 20 -> 40.  Timeout on s390x.

besser82:
* Some more improvements to the rpkg spec file template.

besser82:
17 Days open now and no reactions from others…

@zackw If there are no other improvements needed nor desired from you in this PR, I'm going to merge this tomorrow at about 10:30 UTC.  From my POV we can fix things and do further development in separate PRs then.

ldv-alt:
On Mon, Feb 03, 2020 at 08:15:36AM -0800, Björn Esser wrote:
> 17 Days open now and no reactions from others…

I'm definitely not a big fan of meson, and introduction of meson as a hard
build requirement for such basic package as libxcrypt will create a build loop
that would have to be broken somehow to re-enable bootstrap.

If you are going to change the build system in a radical way like this,
please make sure that the binaries produced by the build system in all
supported configurations before and after the conversion are unchanged.

zackw:
@ldv-alt 

> I'm definitely not a big fan of meson, and introduction of meson as a hard build requirement for such basic package as libxcrypt will create a build loop that would have to be broken somehow to re-enable bootstrap.

That sounds like a pretty solid reason not to do this. Also, I'm actually less happy with meson now than I was when I started working on this PR - specifically, I keep tripping over bugs that have been languishing in the meson issues list for years now, sometimes with a patch attached ... which is rather similar to the situation with autoconf.  So far, there always has been a workaround, but if we're not getting away from the land of endless local workarounds with this, it's hard for me to argue that the expanded bootstrap set would be worth it on that basis.

__However.__

Do you have any ideas for how we could fix #96 without switching away from libtool?  The _other_ really compelling reason (in my book) to stop using autotools is that, as far as I can tell, it is _impossible_ to get libtool to generate a library with the soname `libcrypt.so.1.1`, short of modifying the `libtool` script after it's generated, which is a little too fragile for my taste.

Also, how would you feel about a hard build requirement for Python >=3.6 but _not_ for meson+ninja?  That is, we take the part of this patch where I rewrote all the awk scripts in Python, but we keep driving them from an auto\* build system.  That would enable a bunch of nice things in and of itself.  (Note that glibc itself has a hard build requirement for Python >=3.4, as of 2.31, so I would like to think this isn't too much of a stretch.)

rfc1036:
Since I have opened #96: I would rather keep libxcrypt as is than switch to Meson to fix it. Alpha and IA64 are dead architectures and I have a good enough solution for them.
I do not love Python, but probably that dependency would not be a problem as long as the host Python is used when cross-compiling.

ldv-alt:
On Mon, Feb 03, 2020 at 04:24:06PM -0800, Zack Weinberg wrote:
> Do you have any ideas for how we could fix #96 without switching away from libtool?  The _other_ really compelling reason (in my book) to stop using autotools is that, as far as I can tell, it is _impossible_ to get libtool to generate a library with the soname `libcrypt.so.1.1`, short of modifying the `libtool` script after it's generated, which is a little too fragile for my taste.

I don't know of any way to make libtool generate a custom soname besides
hacking the generated libtool script.

> Also, how would you feel about a hard build requirement for Python >=3.6 but _not_ for meson+ninja?  That is, we take the part of this patch where I rewrote all the awk scripts in Python, but we keep driving them from an auto\* build system.  That would enable a bunch of nice things in and of itself.  (Note that glibc itself has a hard build requirement for Python >=3.4, as of 2.31, so I would like to think this isn't too much of a stretch.)

Since python is required to build glibc nowadays, there should be no issues
if python was also required to build libxcrypt.

besser82:
> I'm definitely not a big fan of meson, and introduction of meson as a hard build requirement for such basic package as libxcrypt will create a build loop that would have to be broken somehow to re-enable bootstrap.

@ldv-alt Can you please elaborate a bit more about the build loop?

Don't get me wrong, but as far as I can see, building with Meson doesn't have any heavy dependencies, except for Python >= 3.5.2, which would be needed for `glibc` anyways.  `ninja-build` has no build dependencies, except a working compiler toolchain, either.

For initial bootstrapping (e.g. preparing a stage-1 image for new architectures) it is definitely enough to have those tools available on the host-system used for building.

ldv-alt:
On Wed, Feb 05, 2020 at 02:46:36AM -0800, Björn Esser wrote:
> > I'm definitely not a big fan of meson, and introduction of meson as a hard build requirement for such basic package as libxcrypt will create a build loop that would have to be broken somehow to re-enable bootstrap.
> 
> @ldv-alt Can you please elaborate a bit more about the build loop?
> 
> Don't get me wrong, but as far as I can see, building with Meson doesn't have any heavy dependencies, except for Python >= 3.5.2, which would be needed for `glibc` anyways.  `ninja-build` has no build dependencies, except a working compiler toolchain, either.

$ grep ^BuildRequires rawhide/ninja-build.spec 
BuildRequires:  gcc-c++
BuildRequires:  python2-devel
BuildRequires:  python3-devel
BuildRequires:  asciidoc
BuildRequires:  gtest-devel
BuildRequires:  re2c >= 0.11.3
[python2-devel is essentially ifdef'ed]

$ grep ^BuildRequires rawhide/asciidoc.spec 
BuildRequires: python3-devel
BuildRequires: dblatex
BuildRequires: docbook-style-xsl
BuildRequires: graphviz
BuildRequires: libxslt
BuildRequires: lilypond
BuildRequires: source-highlight
BuildRequires: texlive-dvipng-bin
BuildRequires: texlive-dvisvgm-bin
BuildRequires: vim-filesystem
BuildRequires: symlinks
BuildRequires: automake
BuildRequires: autoconf
BuildRequires: ImageMagick

I didn't say it's impossible to break the loop, but something would have
to be done with ninja-build to re-enable bootstrap.

besser82:
@ldv-alt `ninja-build` can build with small dependency footprint applying [this patch](https://gist.github.com/besser82/16a247d2236a6fd37059fee57b8d4453) to its spec file.

The dependency chain will be really lightweight then:

* compiler toolchain and `autotools`
* `re2c` (no deps, besides compiler toolchain) and `python >= 3.6.3` (needed for `glibc` anyways)
* `meson` and `ninja-build`
* `libxcrypt`
* … do next stage …

solardiz:
Small distros like our Owl happen to have autotools anyway because so many packages build-depend on those, but would have to explicitly add meson to keep or add libxcrypt. While Owl development is on hold now and might not be continued, revitalizing Owl and adding libxcrypt is within consideration, and I certainly would rather not add meson. We similarly avoided cmake so far (we even patched the only package that required cmake to build to avoid that requirement). It'd be a pity if we'd have to bite one or both of those extra bullets for just a handful of packages trying to get creative in what build systems they use. With all their drawbacks, autotools are at least something that so many base packages build with.

solardiz:
With autoconf, it's also possible to ship and optionally use a pre-generated configure script, which can be handy in case the distro's packaged autoconf version is somehow not suitable to build a given package. Not the case with meson, right?

besser82:
> With autoconf, it's also possible to ship and optionally use a pre-generated configure script, which can be handy in case the distro's packaged autoconf version is somehow not suitable to build a given package. Not the case with meson, right?

Not directly.  If that would be the only _real_ show-stopper, I can setup some scripts to make something like that possible.  One would need to compile `ninja-build` bundled with the tarball before being able to build `libxcrypt`, tho.

PR #98, opened 2020-01-18
Speed up ka-sunmd5 by skipping most of the test phrases.

zackw:
ka-sunmd5 is slower than the entire rest of the testsuite put together, because the sunmd5 hash is Just That Slow and we have to do extra tests for it to ensure bug-compatibility.  No easy optimizations are possible and the difficult optimizations are not worth the engineering effort for this obsolete hash.  Instead, skip most of the test phrases.  See comments in test/ka-table-gen.py!h_sunmd5 for further explanation.

This didn't help with Travis cycle time as much as I was hoping it would (32 -> 20 min end-to-end; compare e.g. https://travis-ci.org/besser82/libxcrypt/builds/638572508 with https://travis-ci.org/besser82/libxcrypt/builds/638867278).  I'm posting a PR instead of just committing it in hopes someone has a better idea.


zackw:
Since nobody has a better idea yet, I'm going to go ahead and merge this.  We can always revisit this if someone does eventually have a better idea.

PR #102, opened 2020-04-04
Add e2k architecture support

bircoph:
GLIBC_2.0 is the minimal required version on the e2k architecture.

If you are wondering what the e2k arch is, see https://en.wikipedia.org/wiki/Elbrus_2000


besser82:
Thanks!

besser82:
I've tagged a new release `v4.4.16`.

PR #111, opened 2020-08-18
Fix build on (Free)BSD.

### LINKED ISSUE #110: Can't make/gmake on FreeBSD 12.1
Platform: FreeBSD 12.1 64-bit (running inside Vmware Workstation Player 15) - 4 GB RAM
libxcrypt Release: 4.4.16

1. The Installation file says to use 'make' .. but when I try ./configure MAKE="make" it give me errors. 
a. bash ./autogen.sh output : http://bsd.to/cmc4
b. ./configure or ./configure MAKE="make" .. it gives me following error/output: https://bsd.to/nxDe

2. When I use gmake (4.3) instead using ./configure MAKE="gmake" 
b. running ./configure completes and give this output: https://bsd.to/sMXe
c. but breaks when I do gmake: http://bsd.to/SZzH

besser82:
#### Include <unistd.h> without declaring crypt prototypes.
The prototypes of the crypt{,_r}() function and struct crypt_data possibly defined in <unistd.h> need to be renamed as they might be incompatible with our declarations.
We achieve this by defining macros *BEFORE* including <unistd.h> at the top of <crypt-port.h> and undefining them directly afterwards.

Defining these macros *AFTER* <crypt-symbol-vers.h>, which also defines the same macros for symbol-versioning purposes, was included, would lead to a clashing redefinition of this macro, and thus cause our internal symbol-versioning would not work properly anymore.

Also drop any direct inclusion of <unistd.h> in several source files.

Fixes #110.

***

#### Use sigjmp_buf instead of jmp_buf for the sigjmp() function.
This change is made to be conformant with the IEEE Std 1003.1-2001, also known as POSIX 2001, which explicitly requires the function to be called with struct sigjmp_buf as the first parameter.

While glibc defines both data-structures to be identical, other libc implementations may define them individually, and thus, from the POV of the C compiler to be different and/or incompatible types.

### DIFF
--- a/lib/crypt-port.h
+++ b/lib/crypt-port.h
@@ -50,6 +50,23 @@
 #include <sys/param.h>
 #endif
 
+/* The prototypes of the crypt() function and struct crypt_data possibly
+   defined in <unistd.h> need to be renamed as they might be incompatible
+   with our declaration.
+   Defining these macros *AFTER* <crypt-symbol-vers.h>, which also defines
+   the same macros for symbol-versioning purposes, was included, would lead
+   to a clashing redefinition of these macros, and thus cause our internal
+   symbol-versioning would not work properly anymore.  */
+#ifdef HAVE_UNISTD_H
+#define crypt unistd_crypt_is_incompatible
+#define crypt_r unistd_crypt_r_is_incompatible
+#define crypt_data unistd_crypt_data_is_incompatible
+#include <unistd.h>
+#undef crypt
+#undef crypt_r
+#undef crypt_data
+#endif
+
 #ifndef HAVE_SYS_CDEFS_THROW
 #define __THROW /* nothing */
 #endif
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,7 @@ Please send bug reports, questions and suggestions to
 <https://github.com/besser82/libxcrypt/issues>.
 
 Version 4.4.18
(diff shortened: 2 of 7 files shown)


besser82:
@zackw Are you fine with this approach?

besser82:
@zackw Ping?

lib/crypt-port.h
```diff
@@ -50,6 +50,23 @@
 #include <sys/param.h>
 #endif
 
+/* The prototypes of the crypt() function and struct crypt_data possibly
+   defined in <unistd.h> need to be renamed as they might be incompatible
+   with our declaration.
+   Defining these macros *AFTER* <crypt-symbol-vers-h>, which also defines
```
zackw, line 56: Typo: `crypt-symbol-vers.h`.

zackw (commented):
Sorry for the delay, the day job has been very busy these past several months.

I was originally thinking of going in the opposite direction: remove the includes of `crypt.h` and `crypt-symbol-vers.h` from `crypt-port.h`, then ensure that every source file includes _either_ `crypt.h` _or_ `unistd.h`, never both.  But this seems more robust, honestly, and we can always look into that possibility later.

lib/crypt-port.h
besser82: Fixed in rebased commit.

besser82:
Allrighty.  Will merge as soon as Travis goes green.

PR #112, opened 2020-10-10
Fix typos.

ferivoz:
Typos were found with codespell.


besser82:
Thanks!  I will look into #113 within the next few days.

PR #114, opened 2020-11-25
Meson port

Salamandar:
Meson build system is a modern build system fully explicit (meaning there's no `#define` you didn't want) with clear syntax and that uses Ninja instead of Make.
Meson : 586 lines of code for 3 seconds of configure + build time.
Autotools : ~2400 lines of code for 10s of configure and 2s of build time.

I ported everything except the test scripts that expect a `.lib` directory to check symbols.

BTW, I didn't have to edit any source file, this merge request is all new files ;)


zackw:
I regret to say that we already considered this conversion, put a lot of work into it, and ultimately abandoned the idea, because it would cause too many problems for our downstream redistributors.  See the discussion in pull #97.  I'm going to quote what @solardiz said here, since it summarizes the issue as neatly as anything I could come up with:

> Small distros like our Owl happen to have autotools anyway because so many packages build-depend on those, but would have to explicitly add meson to keep or add libxcrypt. [...] We similarly avoided cmake so far (we even patched the only package that required cmake to build to avoid that requirement). It'd be a pity if we'd have to bite one or both of those extra bullets for just a handful of packages trying to get creative in what build systems they use.

Rather than putting any more work into a change we can't accept, I would like to ask that you consider helping me with the [`zack/no-awk-no-libtool`](https://github.com/besser82/libxcrypt/tree/zack/no-awk-no-libtool) branch, in which I rewrote all the awk scripts in Python, and also try to stop using libtool, which is the source of most of the problems we actually have with autotools.

Salamandar:
Well that's sad. But I can understand this for a library this broadly used.
Note that adding Meson doesn't mean removing Autotools (although that would mean supporting 2 build systems, I guess you don't want that). Also, broadly used projects already use Meson and removed autotools : iputils, zstd, lz4, pulseaudio and probably others.
I can check out the awk scripts, I 100% understand that you want to remove that scripting language.

ldv-alt:
On Fri, Nov 27, 2020 at 09:03:24AM -0800, Salamandar wrote:
> Also, broadly used projects already use Meson and removed autotools:
> iputils, zstd, lz4, pulseaudio and probably others.

Sorry, but this is certainly not the case for zstd and lz4.

I just happen to know a bit about build systems of zstd and lz4:
these projects have always been using plain Makefile files,
they did not remove autotools because they didn't use them
in the first place.

Besides standard make, zstd also supports cmake and meson,
and lz4 has meson support in contrib subdirectory.

Salamandar:
> Sorry, but this is certainly not the case for zstd and lz4.

Yeah, sorry, I had read about it some days ago (i'm using buildroot and it uses Meson for those packages) and i wrongly understood it was about upstream (whereas it was about the Buildroot packages). My bad ^^

zackw:
I feel I should clarify that redistributors need libxcrypt to keep its
build dependencies minimal more than they do for a lot of other
software. I don't fully understand the issue, but it sounds like the
heart of it is that libxcrypt provides functionality that used to be
part of the C library proper. So, at least for now, it's getting built
at the same early stage of an architecture bootstrap that the C
library would be, and that means fewer tools are available.

For the same reason, libxcrypt puts unusual demands on its shared
library build -- all the version map stuff. Libtool doesn't handle
that well, but neither did meson, when I tried it. (Look at how much
of a mess the top-level and lib-directory meson.build files are on the
zack/meson branch.)

Issue #115, opened 2020-12-10
Replace awk and shell scripts with Python or Perl?


zackw:
Quite some time ago I rewrote our large and unwieldy collection of build-time shell and awk scripts in Python 3: https://github.com/besser82/libxcrypt/commit/7a2ce7d549293a1ab971fb16f3d65fa5124808da  (Then I got a little overambitious and tried to get rid of libtool as well, and then the day job intervened and I haven't touched the code in months.)

I would like to dust this patch off and merge it, but it has occurred to me that it might make life easier for our downstream redistributors if the language used was Perl instead of Python.  Automake is written in Perl, so that interpreter is already part of the bootstrap set, where Python might not be.

I'm personally more fluent in Python, but I'd be fine with either, all told.

@solardiz @ldv-alt @fweimer @rfc1036 I know you've all worked on redistribution issues in the past, what do you think?

(N.B. The existing Python script `ka-table-gen.py` would not be rewritten in Perl, because as far as I know there is no Perl equivalent of [`passlib`](https://passlib.readthedocs.io/en/stable/).  But that script is not run during a normal build, so it doesn't matter.)

rfc1036:
Thank you for your concern. In Debian/Ubuntu a significant subset of Perl is an Essential package, hence due to its pervasivity in the distribution it will be around since the early bootstrap of a port. Python would probably require a long period of cross-compilation of the package.

solardiz:
@zackw In Owl, we chose not to have Python in the distro at all. This took some effort to untie a few components from Python (e.g., RPM), but luckily there were not too many and the ties were not too tight. Since then, I hear recent glibc requires Python for build? I don't know what we'd have done if we wanted to revitalize and update Owl. I guess needing to bite the Python bullet would be yet another reason not to continue with the project. Anyway, now that's history.

For libxcrypt, if it's only to be built on systems with glibc, I guess Python might be acceptable if glibc has a similar dependency. A question might be which version of Python. You appear to require 3.6+ already? Some still maintained LTS distros might not have 3.6 except in an extra packages repository. glibc doesn't appear to require 3.6+ yet: https://sourceware.org/glibc/wiki/Style_and_Conventions#Python_usage_conventions "Require python-2.7, but be compatible with python-3.4+ (scripts only used by glibc developers may require python 3)".

Should libxcrypt be usable on systems without glibc, though? Perhaps yes. And perhaps that's a reason against Python then.

[...]

PR #118, opened 2021-01-25
Convert build scripts to Perl

zackw:
Discussion in #115 was generally favorable to the idea of converting all the build (and test) scripts from sh+awk to Perl; this PR does that.

Notable immediate improvements include that there is now _one_ parser for `hashes.conf` and for `libcrypt.map`, that all of the `libcrypt.minver.*` files can be rolled into one, and that (IMHO) the code is overall quite a bit easier to read.  My motivation for making this change, however, is to make it easier to change how the build works in the future; for instance, to add more fields to `hashes.conf`, or to generate a `crypt-symbol-versions.h` that can make use of `__attribute__((symver))`.

I selected perl 5.14 as the minimum supported version, trying to keep it well in the past but still relatively convenient to test with.  I'm open to arguments for moving the minimum in either direction -- 5.14 happens to be the oldest version with comprehensive support for Unicode text (specifically, with a fully implemented `use feature "unicode_strings"`), and that's why I picked it, but we do not actually _need_ support for Unicode text; going in the other direction, the only thing I know I _want_ from newer perls is named subroutine arguments ("signatures") and those are still considered experimental as of the very newest version, so it's probably not a great idea to start using them yet.


besser82 (approved):
I'm fine with the change and will merge after I added explicit BuildRequires on `make` and `perl-core`.

haampie:
Can this please be reverted? Perl depends on libcrypt, so this introduces a circular dependency.

PR #119, opened 2021-02-20
lib/alg-gost3411-core.c: fix conversion error seen by sensitive compilers

gwbres:
Some compilers might be sensitive and raise an Warning/error on this line.

Signed-off-by: Guillaume W. Bres <guillaume.bressaix@gmail.com>

### DIFF
--- a/lib/alg-gost3411-2012-core.c
+++ b/lib/alg-gost3411-2012-core.c
@@ -92,7 +92,7 @@ add512(const uint512_u *x, const uint512_u *y, uint512_u *r)
     for (i = 0; i < 64; i++)
     {
         buf = xp[i] + yp[i] + (buf >> 8);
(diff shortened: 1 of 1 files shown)


besser82:
Thank you for the contribution!

gwbres:
@besser82, thank you for merging this little patch!

I added your package into Buildroot, a tool to generate embedded linux images easily.
I personally use your package on the beaglebone, bb-black and the zynq7.
This problem was seen on "mips32" platform though. 

Like you, I prefer having all warnings & errors available when compilation, the build is harder to achieve, the code is supposedly cleaner.
But on the other hand, having -Werror pass on a couple machines (like yours, & the one I tested, does not guarantee it does on others (like mips for instance, who was tacky on this particular "cast").
The Buildroot guys had me disable the -Werror in configure at the moment

PR #126, opened 2021-04-30
build-aux/compute-symver-floor: Fix handling of spaces in CFLAGS

rpurdie:
If you pass CFLAGS with a leading space, " " gets passed to popen and convinces
gcc to try and open a file called " ". This results in a confusing error message
like:

x86_64-pokysdk-linux-gcc: error: : No such file or directory

Avoid this by stripping empty elements out of CFLAGS.

Signed-off-by: Richard Purdie <richard.purdie@linuxfoundation.org>


besser82:
I've applied a slightly modified version of your patch in fed81d1a98fab52839b60a55447d368899034847.

Thanks for your contribution!

PR #128, opened 2021-05-06
Indicate weak hashes by crypt_checksalt().

besser82:
The hashes.conf file used to define the hashing methods supported by libxcrypt defines a group of STRONG hashing methods.  Those methods, that are not members of this group, are considered to be too weak for still being used for new passphrases.

The crypt_checksalt() function now returns `CRYPT_SALT_METHOD_LEGACY` for such settings.

### DIFF
--- a/lib/crypt.c
+++ b/lib/crypt.c
@@ -68,6 +68,7 @@ struct hashfn
   /* The type of this field is unsigned char to ensure that it cannot
      be set larger than the size of an internal buffer in crypt_gensalt_rn.  */
   unsigned char nrbytes;
+  unsigned char is_strong;
 };
 
 static const struct hashfn hash_algorithms[] =
--- a/lib/crypt.c
+++ b/lib/crypt.c
@@ -321,7 +321,12 @@ crypt_checksalt (const char *setting)
   const struct hashfn *h = get_hashfn (setting);
 
   if (h)
-    retval = CRYPT_SALT_OK;
+    {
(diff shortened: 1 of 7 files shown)


besser82:
@zackw Failures on Travis are unrelated.  Are you okay with merging, or are there some changes you'd suggest?

besser82:
@zackw If you have no objections, I'm going to merge this as is during today.

lib/crypt.c
```diff
@@ -320,7 +321,12 @@ crypt_checksalt (const char *setting)
   const struct hashfn *h = get_hashfn (setting);
 
   if (h)
-    retval = CRYPT_SALT_OK;
+    {
+      retval = CRYPT_SALT_OK;
+
+      if (h->is_strong == 0)
+        retval = CRYPT_SALT_METHOD_LEGACY;
+    }
```
zackw, line 329: I would probably have written this as

```
if (h)
  retval = h->is_strong ? CRYPT_SALT_OK : CRYPT_SALT_METHOD_LEGACY;
```

but I don't insist on that change.

zackw:
Apart from what I just annotated the code with, LGTM.  Sorry for the delay, I was busy wrapping up the semester at my university.

lib/crypt.c
besser82: Yeah, I didn't use the ternary operator here on purpose, as we would likely need to change to the style I used as soon as we start implementing one of the other return codes.

PR #131, opened 2021-06-12
Setup github actions CI

besser82:
As the free TravisCI is going to pass away in near future, we should adopt our CI to github actions.

PR #136, opened 2021-07-21
Update codecov.yml to codecov/codecov-action@v2.

zackw:
codecov-action@v1 is deprecated due to its inextricable coupling with
Codecov’s shell-script uploader, which is also deprecated, due to its
being an unmaintainable pile of shell script.  See
https://github.com/codecov/codecov-action/blob/635d4e88ad8c1f75b56277efd9ec186c3359727f/README.md#%EF%B8%8F--deprecration-of-v1 [sic]
and https://about.codecov.io/blog/introducing-codecovs-new-uploader/
for gory details.

The new uploader does not support the ‘gcov_executable‘ and
‘gcov_path_exclude‘ parameters yet, so for the time being we go back
to preprocessing the coverage data with lcov (as we were doing on
Travis).

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -54,10 +54,15 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v2
-      # note: the 'llvm' package is needed for llvm-cov (only when using clang)
+
     - name: Install packages
-      if: ${{ matrix.compiler == 'clang' }}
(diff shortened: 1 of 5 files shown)


zackw:
It appears that this also gets us branch tracking back and that's why the apparent coverage loss.

I would like to merge this ahead of #131 to decouple them.

besser82 (approved):

Issue #135, opened 2021-07-07
Make generic code, hashing methods, and documentation agree re forbidden characters in hashes


zackw:
`crypt.5` currently says

> Hashed passphrases are always entirely printable ASCII, and do not contain any whitespace or the characters ‘:’, ‘;’, ‘*’, ‘!’, or ‘\’.

However, our generic code [currently only enforces the absence of ‘:’ and newline](https://github.com/besser82/libxcrypt/blob/ff1bfae58b5fa9bfd53f6acdaffcd09274ad6fc6/lib/crypt.c#L131).  Prior to 2112d2bb95f294846df61180003f079078a3f83d, there was no generic check at all.  Prior to b2b813adbd9d7cc7d20a9b984ef27b64df8c15d9, some but not all hashing methods implemented much more restrictive rules (allowing only their own base64 alphabet in the salt part of a setting); these were found to conflict with existing salts (see #105).  The setting-parsing code in each hashing method is ad hoc and difficult to audit for what syntax it actually accepts; there might be other quirks hiding.

I think the rule as stated in the documentation is abstractly correct, and I would _like_ to make the generic code enforce that rule and then remove the redundant checks in the hashing methods.  However, I'm worried about this breaking existing hashes as the pre-b2b813adbd9d7cc7d20a9b984ef27b64df8c15d9 code did.  I'm soliciting comments from the community.

OleHolmNielsen:
Would the POSIX.1-2017 standard for the crypt() function be the best choice for characters?  POSIX declares in
https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/crypt.html that only the following characters may be used in the salt:
```
The salt argument shall be a string of at least two bytes in length not including the null character chosen from the set:
a b c d e f g h i j k l m n o p q r s t u v w x y z
A B C D E F G H I J K L M N O P Q R S T U V W X Y Z
0 1 2 3 4 5 6 7 8 9 . / 
```

besser82:
Basically there are several RFCs defining base64 encoding tables for different applications.  And there is the POSIX base64 table for MCF.

If I compare them there are just a few different characters allowed in them: `+-_,=`, besides the POSIX MCF table.

Then we have two known ways to consume those salt characters by the hash methods:

1. Just use them as a c-string for computation, e.g. sha{256,512}crypt and md5crypt.
2. Translate them back to the original raw bytes and use those for computation, e.g. yescrypt.

[...]

PR #138, opened 2021-08-05
Stricter checking of invalid salt characters.

besser82:
This PR is a proposal for implementing the restrictions as discussed in #135.

***

For now we consider the following characters to be valid for any salt string in the generic check:

```
0 1 2 3 4 5 6 7 8 9
a b c d e f g h i j k l m n o p q r s t u v w x y z
A B C D E F G H I J K L M N O P Q R S T U V W X Y Z
. / + - _ = , $ " # % & ' ( ) < > ? @ [ ] ^ ` { | } ~
```


lib/crypt-port.h
```diff
@@ -470,6 +471,10 @@ make_failure_token (const char *setting, char *output, int size);
    because it's used as a C string in a few places.  */
 extern const unsigned char ascii64[65];
 
+/* The table with invalid salt characters used for inital checking
+   of salt strings.  */
+extern const unsigned char badsalt_chars[166];
```
solardiz, line 476: Typo: "inital". I don't think you gain anything by including the NUL explicitly and using hard-coded array size (and I didn't verify the 166 is right). I'd just use `[]` here and let the compiler add the NUL.

lib/crypt.c
```diff
@@ -110,6 +110,15 @@ get_hashfn (const char *setting)
   return 0;
 }
 
+static int
+check_badsalt_chars (const char *setting, size_t set_size)
+{
+  if (strcspn (setting, (const char *) badsalt_chars) != set_size)
+    return 1;
+
+  return 0;
```
solardiz, line 119: I think the explicit returns of 1 and 0 don't improve readability for anyone familiar with C. I suggest you return the result of the comparison directly, making this function a one-liner.

lib/util-badsalt-chars.c
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+/* Table of characters that are invalid for MFC salt-strings.  */
+
+#include "crypt-port.h"
+
+const unsigned char badsalt_chars[166] =
+  "\x01\x02\x03\x04\x05\x06\x07\x08\x0a\x0b\x0c\x0d\x0e\x0f"
+  "\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f"
+  " !*:;\\\x7f"
```
solardiz, line 26: So you block a few more printable characters here. The previous list of just two characters to block had solid rationale: those characters wouldn't work right in an `/etc/shadow` entry _anyway_. Now you're blocking characters that would actually work inside a salt. I can see how those are potentially somewhat problematic, but I am not convinced this is obviously the right thing to do. I don't object, though.

For the control characters and the 8-bit range, they're so numerous that maybe checking for them with `strcspn` would have measurable performance impact. Not high and usually negligible, but measurable. With so much effort put into optimizing some of the password hashing implementations, it can be a bit silly to use inefficient constructs in implementing the API. So, while I like how this patch doesn't change the code too much (only replaces the string constant and moves the check inside a function and uses that function twice), maybe it'd be more appropriate to be blocking below 0x20 and above 0x7e with `if` inside a loop. Or choose to not block those.

lib/crypt-port.h
besser82: Addressed in rebased commit.

lib/crypt.c
besser82: Addressed in rebased commit.

lib/util-badsalt-chars.c
besser82: Addressed in rebased commit.

besser82:
@solardiz Should be fine now.

besser82:
@zackw You can ignore the one failing test, as it failed for $foo and not the code changes.

lib/crypt.c
```diff
@@ -110,6 +110,18 @@ get_hashfn (const char *setting)
   return 0;
 }
 
+static int
+check_badsalt_chars (const char *setting)
+{
+  size_t i = 0;
```
solardiz, line 116: Redundant initialization. I'd keep one inside `for` only.

lib/crypt.c
```diff
@@ -110,6 +110,18 @@ get_hashfn (const char *setting)
...
+
+  for (i = 0; setting[i] != '\0'; i++)
+    if (setting[i] <= 0x20 || setting[i] >= 0x7f)
```
solardiz, line 119: Signed comparisons (except on the few architectures where `char` is unsigned by default). They will work OK in practice, but non-intuitively the 8-bit chars will be caught by the first check and not by the second. I don't have a preference on leaving this as-is vs. adding extra casts vs. adding a comment.

lib/crypt.c
```diff
@@ -138,10 +150,17 @@ do_crypt (const char *phrase, const char *setting, struct crypt_data *data)
...
-     <https://github.com/besser82/libxcrypt/issues/105>.  */
-  if (strcspn (setting, ":\n") != set_size)
+     enforced this rule, so for now we are being cautious and consider
+     the following characters to be valid for any salt string in this
+     generic check:
+
+         0 1 2 3 4 5 6 7 8 9
+         a b c d e f g h i j k l m n o p q r s t u v w x y z
+         A B C D E F G H I J K L M N O P Q R S T U V W X Y Z
+         . / + - _ = , $ " # % & ' ( ) < > ? @ [ ] ^ ` { | } ~
```
solardiz, line 160: FWIW, I didn't verify that this indeed the exact list of characters not included among the bad ones.

lib/crypt.c
```diff
@@ -334,8 +353,9 @@ crypt_checksalt (const char *setting)
 {
   int retval = CRYPT_SALT_INVALID;
 
-  if (!setting ||          /* NULL string */
-      setting[0] == '\0')  /* empty passphrase, or descrypt without salt */
+  if (!setting ||            /* NULL string */
+      setting[0] == '\0' ||  /* empty passphrase, or descrypt without salt */
+      check_badsalt_chars (setting))  /* bad salt chars */
     goto end;
```
solardiz, line 359: FWIW, I don't understand what "descrypt without salt" means in the comment here (unchanged in this PR). I think there's no such thing as "descrypt without salt".

lib/crypt.c
besser82: Fixed in rebased commit.

lib/crypt.c
besser82: Fixed in rebased commit.

lib/crypt.c
besser82: Fixed in rebased commit.

lib/crypt.c
besser82: This is verified by `test/special-char-salt.c`.

solardiz (approved):
Looks OK to me now.

besser82:
> Looks OK to me now.

Thank you for your time and review!

besser82:
@zackw Do you have any further remarks or objections on this?

zackw:
I’m unfortunately not going to be able to look at this until tomorrow.

zackw (approved):
LGTM except for this comment in crypt.c (https://github.com/besser82/libxcrypt/blob/topic/besser82/badsalt_chars/lib/crypt.c#L144-L163)

> ```
> ... In principle this check ought to match the rule
>     stated in crypt(5):
>         Hashed passphrases are always entirely printable ASCII, and
>         do not contain any whitespace or the characters ':', ';',
>         '*', '!', or '\\'.  (These characters are used as delimiters
>         and special markers in the passwd(5) and shadow(5) files.)"
>     However, individual hash function implementations have not always
>     enforced this rule, so for now we are being cautious and consider
>     the following characters to be valid for any salt string in this
>     generic check:
>         0 1 2 3 4 5 6 7 8 9
>         a b c d e f g h i j k l m n o p q r s t u v w x y z
>         A B C D E F G H I J K L M N O P Q R S T U V W X Y Z
>         . / + - _ = , $ " # % & ' ( ) < > ? @ [ ] ^ ` { | } ~
> ```

The issue is that "consider[ing] [only] the following characters to be valid" is in fact the same rule as "always printable ASCII and do not contain `: ; * ! \\`", so we _aren't_ being cautious relative to the documentation anymore.  I suggest the following revision:

```
/* Check a setting string for generic validity, according to the rule
   stated in crypt(5):

      "Hashed passphrases are always entirely printable ASCII, and do
      not contain any whitespace or the characters ':', ';', '*', '!',
      or '\\'.  (These characters are used as delimiters and special
      markers in the passwd(5) and shadow(5) files.)"

   There is a precautionary case for rejecting additional ASCII
   punctuation, particularly other characters often given syntactic
   significance in configuration files, such as " ' and #.  However,
   this check didn't used to exist at all, and some of the hash
   function implementations don't restrict the set of byte values they
   they will accept in their setting strings (particularly in the salt
   component) either.  Thus, to maintain compatibility with the widest
   variety of existing hashed passphrases, we are only enforcing the
   documented rule for now.

   See <https://github.com/besser82/libxcrypt/issues/135> for
   additional discussion.  */
```

I also suggest moving this comment to above the definition of check_badsalt_chars.

besser82:
Fixed.  Let's wait for the CI and I'll merge on cli then.

Issue #142, opened 2021-10-22
Building with no symbol versioning is broken


galaxy4public:
TL;DR -- `./configure --disable-symvers` produces a configuration one cannot actually build since it will fail during the "make" phase.

Although `./configure` provides an option to disable symbol versioning (i.e. `--disable-symvers`) it does not really work since there are multiple places in the source code where symbol versioning is engaged.  For starters, the `./build-aux/scripts/gen-crypt-symbol-vers-h` script produces a header with versioning info.  I am building this library on a musl-based system and musl does not support versioning (well, it can work with it, but I try to keep it clean, so don't want to have any symvers crawling into the libraries).  Perhaps, I will submit a patch for this issue soon.

PR #143, opened 2021-10-28
Provide an opportunity to build libxcrypt without symvers

galaxy4public:
This tries to (partially) address #142 .  This PR introduces a condition inside the `gen-crypt-symbol-vers-h` helper script, which is used to generate a header file with versioning info.  If the `DO_NOT_GENERATE_SYMVERS` environment variable is present, then the script won't produce versioning info.

I think maintainers have a better idea on how to integrate this into the autotools build system than I am.  Ideally, when somebody passes `--disable-symvers` to configure the aforementioned environment variable should be set before calling the script.


besser82:
I've come up with a more elaborated solution to this in #159.

PR #146, opened 2021-11-30
Fixes for issue 145

besser82:
doc/crypt.5: Document the recommended amount of salt-bits for `yescrypt`.

Also document the same value as the recommended amount for `gost-yescrypt` and `scrypt`.

***

lib: Silently truncate rbytes after a maximum of 512 bits for `yescrypt`.

Likewise for `gost-yescrypt` and `scrypt`, as those hashing methods share the same codebase.


besser82:
@solardiz Please let me know, if you are fine with these changes.

PR #147, opened 2022-01-05
Generate a sufficiently random stop-commands token

quapka:
The token in ::stop-commands::{token} GA workflow command has to be a cryptographically random and secure string. Definitely not a hardcoded string literal. See the docs for details: https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#stopping-and-starting-workflow-commands

I've enabled my local GitHub Actions to see if I can see the changes in action, however, only some of the Actions ran and the call to `ci-log-logfiles` has not been triggered.

### DIFF
--- a/build-aux/ci/ci-log-logfiles
+++ b/build-aux/ci/ci-log-logfiles
@@ -1,12 +1,18 @@
 #! /bin/sh
 
+generate_secure_token() {
+    echo head --bytes 64 /dev/urandom | sha512sum --binary | cut --fields 1 --delimiter ' '
+}
+
 dump_log () {
     if [ -s "$1" ]; then
+        local stop_command_token="$(generate_secure_token)"
         echo "::group::$1"
-        echo '::stop-commands::resume-50YEO1zJ8HSXH4Zy'
+        echo "::stop-commands::$stop_command_token"
         cat "$1"
-        echo '::resume-50YEO1zJ8HSXH4Zy::'
+        echo "::$stop_command_token::"
         echo '::endgroup::'
+        unset stop_command_token
     fi
 }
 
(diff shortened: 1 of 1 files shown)


besser82:
Thanks for your contribution!

quapka:
> Thanks for your contribution!

My pleasure.

I've been thinking of directly producing random tokens using:
```bash
n_bytes=64 # use variable for sake of documentation
xxd -len "$n_bytes" -plain -cols "$n_bytes" /dev/urandom
```

While good to know I don't think this needs a revisit. The current fix is closer to the suggestion in the docs.

besser82:
> > Thanks for your contribution!
> 
> My pleasure.
> 
> I've been thinking of directly producing random tokens using:
> 
> ```shell
> n_bytes=64 # use variable for sake of documentation
> xxd -len "$n_bytes" -plain -cols "$n_bytes" /dev/urandom
> ```
> 
> While good to know I don't think this needs a revisit. The current fix is closer to the suggestion in the docs.

I personally prefer the solution as merged, since it uses basic system command reliably present by default on any very basic UNIX(-alike) installation, while the `xxd` command requires some parts of `vim`.

quapka:
Oh, my bad. I'm definitely for a more cross-platform solution in this case. The additional processes are no a big issue and the fix is readable as is anyway.

besser82:
> Oh, my bad. I'm definitely for a more cross-platform solution in this case. The additional processes are no a big issue and the fix is readable as is anyway.

Don't mind.  Both solutions would do the same, it's just a matter of the target audiance; using system commands is prefered in CI environments, where more sophisticated solutions suit a developer's workstation.

quapka:
Yes, I see it the same. Just today-i-learned about the connection between `xxd` and `Vim`.

PR #148, opened 2022-01-05
Handle properly looping over filenames with whitespace

quapka:
Using `-exec` would be another option, maybe slightly simpler.


besser82:
LGTM!

PR #154, opened 2022-04-19
libcrypt.minver: Mark LoongArch as free of legacy ABIs

xen0n:
This architecture is new, so there is no existing binaries to consider
legacy compatibility for.

### DIFF
--- a/lib/libcrypt.minver
+++ b/lib/libcrypt.minver
@@ -60,6 +60,7 @@ GLIBC_2.3    kfreebsd.*gnu   i[3-9]86
 ERROR        kfreebsd.*gnu   .
 
 # Linux with GNU libc
+XCRYPT_2.0   linux.*gnu      loongarch
 GLIBC_2.35   linux.*gnu      or1k
 GLIBC_2.33   linux.*gnu      riscv32
 GLIBC_2.32   linux.*gnu      arc


zhangn1985:
please consider debian, it still uses legacy API.

thesamesam:
> please consider debian, it still uses legacy API.

Debian started enabling the legacy API unnecessarily for an arch where no legacy binaries existed? Why?

xen0n:
> please consider debian, it still uses legacy API.

Debian upstream doesn't support LoongArch yet, so you actually have the chance to do the right thing and just get rid of the "compat" stuff, after all the LoongArch glibc port arrived *after* the crypt support in glibc was deprecated.

We especially can't do anything as ling as your "debian" work isn't public.

besser82:
@xen0n, thank you for your contribution!

This looks wrong to me, as LoongArch LP64 is well supported by glibc 2.36 and its yet not officially deprecated `libcrypt`.

besser82:
This issue has been addressed with  [release v4.4.29](https://github.com/besser82/libxcrypt/releases/tag/v4.4.29).

xen0n:
> @xen0n, thank you for your contribution!
> 
> This looks wrong to me, as LoongArch LP64 is well supported by glibc 2.36 and its yet not officially deprecated `libcrypt`.

This is a bit unfortunate, admittedly I didn't watch as closely as I'd want to when they upstreamed the glibc port, assuming the glibc libcrypt implementation was universally deprecated long ago. And it's still the case that no LoongArch distribution exists where `libcrypt.so.1` usage cannot be removed without significant breakage; on the contrary the Gentoo loong port has completely eliminated it. So, marking the compat symbol version as `GLIBC_2.36` actually *breaks* Gentoo, and IMO we should instead deprecate the glibc libcrypt for LoongArch ASAP and make the few Loongson-maintained distros follow suit. After all, there are very few new world users compared to other more mature arches, and the Loongson distros enabling the glibc libcrypt are AFAIK not fully open-source so we can't meaningfully support them anyway.

PR #157, opened 2022-10-01
restore ucontext api functionality check

yu-re-ka:
In c3f01c72b303cbbb0cc8983120677edee2f3fa4b the use of the ucontext api in the main program was removed, and with it the configure check for it. However, the ucontext api is still used in the "explicit_bzero" test and this has the same problem as described in the comment:

> The ucontext.h functions that we use were withdrawn from
> POSIX.1-2008, so the existence of the header does not prove
> we can use the functions.

Thus, restore the full configure check and use it instead of the header check for the explicit_bzero test.

See https://bugs.gentoo.org/838172

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -99,7 +99,6 @@ fi
 AC_CHECK_HEADERS_ONCE([
   fcntl.h
   stdbool.h
-  ucontext.h
   sys/cdefs.h
   sys/random.h
   sys/syscall.h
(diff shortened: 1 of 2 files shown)


besser82:
Fixed in https://github.com/besser82/libxcrypt/commit/13435fdc1b8c704fd46b0a9027791db0134bd77b.

PR #160, opened 2022-11-07
Fix warning about truncating conversion

moinakb001:
Unnecessary warning when compiling with AVX support leads to build failures with Werror. Simple cleanup.


besser82:
LGTM!  Thank you!

PR #167, opened 2023-02-03
Fix for compilation on Windows

brechtsanders:
This fix allows the library to build on Windows (at least with MinGW-w64).

`.symver` is only supported for ELF format but Windows uses COFF/PE.

Workaround dummy define of `symver_set()`


besser82:
Thank you for your contribution!

PR #171, opened 2023-06-06
Make BuildCommon.pm compatible with latest perl

### LINKED ISSUE #170: Fails to configure with Perl 5.38.0(-RC0)
https://github.com/Perl/perl5/blob/blead/pod/perldelta.pod#changes-to-existing-diagnostics
^ The "when" keyword moves from "experimental" to "deprecated in Perl 5.38. 

This means it emits a warning, which triggers the `use warnings FATAL => 'all';` at the start of BuildCommon.pm.

```
checking for valgrind... no
checking for Python 3.>=6 with Passlib... not found
configure: Disabling the "regen-ka-table" target, missing Python requirements.
checking whether all ucontext.h functions are available... yes
when is deprecated at /var/tmp/portage/sys-libs/libxcrypt-4.4.34/work/libxcrypt-4.4.34/build-aux/scripts/BuildCommon.pm line 522.
Compilation failed in require at /var/tmp/portage/sys-libs/libxcrypt-4.4.34/work/libxcrypt-4.4.34/build-aux/scripts/expand-selected-hashes line 28.
BEGIN failed--compilation aborted at /var/tmp/portage/sys-libs/libxcrypt-4.4.34/work/libxcrypt-4.4.34/build-aux/scripts/expand-selected-hashes line 28.
configure: error: bad value 'all' for --enable-hashes
```

Leont:
It was previously using an experimental feature that has since been dropped. This removes the use of that feature.

This fixes #170

### DIFF
--- a/build-aux/scripts/BuildCommon.pm
+++ b/build-aux/scripts/BuildCommon.pm
@@ -11,7 +11,6 @@ use v5.14;    # implicit use strict, use feature ':5.14'
 use warnings FATAL => 'all';
 use utf8;
 use open qw(:utf8);
-no  if $] >= 5.018, warnings => 'experimental::smartmatch';
 no  if $] >= 5.022, warnings => 'experimental::re_strict';
 use if $] >= 5.022, re       => 'strict';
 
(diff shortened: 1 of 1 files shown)


besser82:
Thank you for your contribution!  I will merge on cli as soon as all tests are finished.

PR #172, opened 2023-06-24
Remove smartmatch usage from gen-crypt-h

akhuettel:
This is also needed for building with Perl 5.38... another smartmatch usage that now triggers a deprecation warning.

BTW, it's really a bad idea to treat these warnings all as errors. Otherwise we'd get a warning now, but not a failure... more time to fix stuff...


thesamesam:
@besser82 Could you take a look? Perl 5.38 is now out. Cheers!

besser82:
LGTM!  Thank you!

PR #109, opened 2020-08-13
Add the needed bits for compiling with link-time optimization.

besser82:
GCC 10.2 and LLVM/Clang offer initial support for building libraries, that are using symbol versioning features, with LTO.

To make use of this, the exported versioned symbols need their visibility attributes declared explicitly.

Fixes #24.

Issue #24, opened 2018-08-04
lto linking, missing symbols


asavah:
gcc 8.2.0
binutils 2.31.1
glibc 2.28 (with --disable-crypt)

crosscompilation for aarch64, libxcrypt v4.1.1
configure --enable-shared --disable static

If libxcrypt is built with lto in *flags almost all symbols are missing from the library

lto build - bad
```
ls libcrypt.so.1.1.0 -ah
-rwxr-xr-x 1 asavah asavah 9496 Aug  5 00:46 libcrypt.so.1.1.0

nm -DC libcrypt.so.1.1.0
                 w __cxa_finalize
0000000000000000 A GLIBC_2.17
                 w __gmon_start__
                 w _ITM_deregisterTMCloneTable
                 w _ITM_registerTMCloneTable
0000000000000000 A OW_CRYPT_1.0
0000000000000000 A XCRYPT_2.0
```

Non lto build (gold linker) - good
```
ls libcrypt.so.1.1.0 -al
-rwxr-xr-x 1 asavah asavah 146864 Aug  5 00:49 libcrypt.so.1.1.0


nm -DC libcrypt.so.1
                 U __assert_fail
                 U close
000000000000a470 T crypt
000000000000a470 T crypt
0000000000008390 T crypt_gensalt
0000000000008390 T crypt_gensalt
0000000000008390 T crypt_gensalt
0000000000005530 T crypt_gensalt_ra
0000000000005530 T crypt_gensalt_ra
0000000000005530 T crypt_gensalt_ra
00000000000053b0 T crypt_gensalt_rn
00000000000053b0 T crypt_gensalt_rn
00000000000053b0 T crypt_gensalt_rn
0000000000005350 T crypt_r
0000000000005350 T crypt_r
0000000000005250 T crypt_ra
0000000000005250 T crypt_ra
00000000000051c0 T crypt_rn
00000000000051c0 T crypt_rn
                 w __cxa_finalize
000000000000aed0 T encrypt
000000000000aeb0 T encrypt_r
                 U __errno_location
                 U explicit_bzero
000000000000a470 T fcrypt
                 U free
                 U getentropy
                 U getrandom
0000000000000000 A GLIBC_2.17
                 w __gmon_start__
                 w _ITM_deregisterTMCloneTable
                 w _ITM_registerTMCloneTable
                 U malloc
                 U memcmp
                 U memcpy
                 U memset
                 U open
0000000000000000 A OW_CRYPT_1.0
[...]

PR #176, opened 2023-10-07
doc: Update LTO support status in README.md

xry111:


zackw:
With the text as is, it would not be correct to merge this patch until after #109 is merged.  The necessary _compiler_ support now exists, but libxcrypt must also be changed to _use_ that support.

@xry111 If you would like to revise your change so that it accurately summarizes the present state of affairs (see #24 as well as #109), then I'll consider merging it.

xry111:
> With the text as is, it would not be correct to merge this patch until after #109 is merged. The necessary _compiler_ support now exists, but libxcrypt must also be changed to _use_ that support before we can say "LTO is supported".
> 
> @xry111 If you would like to revise your change so that it accurately summarizes the present state of affairs (see #24 as well as #109), then I'll consider merging it.

We are already using that support since 38f7f06ec358561a01636f0d990ed98051b9120a, AFAIK.  But #109 already contains the wanted README.md change I'll close this.

PR #178, opened 2023-12-19
configure: Only text the makecontext signature we need

fweimer-rh:
The `test/explicit-bzero.c` test uses a start routine without any arguments.  There is no need for the multi-argument version.

This avoids a build failure with glibc and future compilers. The GNU C library declares the `makecontext` callback of type `void (*) (void)`, so no cast is needed.  On other systems, the type may be the (currently distinct) type `void (*) ()`, but given that this only affects the ability to execute a test, no further machinery is added here to detect that different type.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -303,13 +303,11 @@ AS_IF([test $ac_cv_header_ucontext_h = yes],
 #include <ucontext.h>
 static int x;
 static void fn1(void) {}
-static void fn2(int a, int b) { x = a - b; }
 ]], [[
   ucontext_t uc1, uc2;
   if (getcontext(&uc1)) return 1;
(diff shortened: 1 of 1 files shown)


neverpanic:
There's a typo in the commit message ("Only text …" instead of "Only test …"). Maybe whoever ends up merging this can amend the commit and fix that before merging.

besser82:
Thank you, @fweimer-rh!

I've fixed the commit message as mentioned by @neverpanic and commited in 2572bd2553df79d326af01c3536eeb523f962fe6.

PR #185, opened 2024-03-27
Man page edits

### LINKED ISSUE #29: crypt.5 .hash's reporting of effective hash size is wrong
As discussed in https://github.com/besser82/libxcrypt/pull/27#discussion_r213801730, the `.hash` macro has been changed somewhere between my revision in Owl (and its later life in ALT Linux?) and libxcrypt's to change the reporting for traditional DES-based hashes from:

```
       Effective key size
              up to 56 bits

       Hash size
              64 bits
```

to:

```
       Hash size
              64 bits (effectively 56)
```

(There are other differences as well, but those look OK.)

The above change is wrong because a limit on key size doesn't reduce the effective hash size as claimed.

First, a cryptographic 56-bit key to 56-bit hash function would likely produce far more collisions (over its full 56-bit key space) than a cryptographic 56-bit to 64-bit hash function is expected to. If this isn't immediately obvious, please consider extreme cases such as a 1-bit to 1-bit vs. a 1-bit to 64-bit hash function (in the former case, 2 out of 4 possible functions would produce collisions half of the time, whereas in the latter an extremely low percentage of the possible functions would produce collisions), or an 8-bit to 8-bit (expect only ~162 different outputs over the 256 input keys space) vs. an 8-bit to 64-bit hash function (expect no collisions).

Second, even with the key input limited to 56 bits, almost the full 2^64 variety of 64-bit hash outputs are likely reachable via varying key+salt inputs.

The same issue is probably also present for BSDI-style DES-based hashes, where the password input length is not limited to 8 characters, but internally the key size is effectively limited to 56 bits, creating a "narrow pipe" within the hash function (except for the salt input, which is also treated separately). This "narrow pipe" is similarly _not_ the same as an effectively reduced hash output size.

Perhaps this change to `.hash` should be reverted or revised, and some uses of `.hash` may need to be revised accordingly. I understand that the lines:

```
       Effective key size
              limited by the hash size only
```

as would be reported by my original `.hash` for most hash types didn't look pretty. Maybe we can have the macro omit this reporting for hashes where this is the case, only leaving the reporting for the few special-case hashes?

solardiz:
This is motivated by the criticism in https://www.openwall.com/lists/yescrypt/2024/03/21/1 that our wording "CPU time cost" was misleading, given that with memory-hard methods it also affects memory usage. That's addressed in the first commit here. The second commit is based on my proof-reading of the entire `crypt.5`.

Not yet changed here, but something we might also want to change, is renaming of the `count` parameter to `cost`. Perhaps not only in man pages, but also in code.

Edit: This also fixes #29

### DIFF
--- a/doc/crypt.3
+++ b/doc/crypt.3
@@ -278,6 +278,9 @@ below.)
 All four functions set
 .Va errno
 when they fail.
+When the functions succeed, the value of
+.Va errno
+is unspecified and must not be relied upon.
 .Sh ERRORS
 .Bl -tag -width Er
 .It Er EINVAL
--- a/doc/crypt.3
+++ b/doc/crypt.3
@@ -243,7 +243,7 @@ and
 .Nm crypt_ra
 write an
 .Em invalid
-hashed passphrase to the
+hash to the
 .Fa output
 field of their
 .Fa data
@@ -350,7 +350,7 @@ returned by
 The behavior described above for this implementation,
 setting
 .Va errno
-and returning an invalid hashed passphrase different from
+and returning an invalid hash different from
 .Fa setting ,
 is chosen to make these applications fail closed when an error occurs.
 .Pp
--- a/doc/crypt.5
+++ b/doc/crypt.5
@@ -33,7 +33,7 @@ However, with a strong hashing method,
 guessing will be too slow for the attacker
 to discover a strong passphrase.
 .Pp
-All of the hashing methods use a
+Most of the hashing methods use a
 .Dq salt
 to perturb the hash function,
 so that the same passphrase may produce many possible hashes.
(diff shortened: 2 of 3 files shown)


solardiz:
This also fixes #29

gurucubano (commented):
The changes are fine with me.

solardiz:
@vt-alt You might want to review my edits of the `gost-yescrypt` description here.

vt-alt:
@solardiz LGTM

solardiz:
@besser82 @zackw Can you please merge this PR? I see no reason not to. You may say that more changes are desired, but that's no reason not to merge the changes so far. Thanks!

zackw:
I have not had time to look at these changes _at all_.  Please be patient.

PR #188, opened 2024-07-10
Add sm3crypt backend

uudiin:
Add OSCCA SM3 secure hash (OSCCA GM/T 0004-2012 SM3) backend and tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,7 @@ test/alg-pbkdf-hmac-sha256
 test/alg-sha1
 test/alg-sha256
 test/alg-sha512
+test/alg-sm3
 test/alg-yescrypt
 test/badsalt
 test/badsetting
--- a/Makefile.am
+++ b/Makefile.am
@@ -80,6 +80,7 @@ noinst_HEADERS = \
 	lib/alg-sha1.h \
 	lib/alg-sha256.h \
 	lib/alg-sha512.h \
+	lib/alg-sm3.h \
 	lib/alg-yescrypt.h \
 	lib/byteorder.h \
 	lib/crypt-obsolete.h \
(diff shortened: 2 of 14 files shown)


solardiz:
Are these hashes already in use? If not yet, you could want to take the same approach as with GOST (see gost-yescrypt here), not the same as with SHA-crypt (which you did in this PR now). I guess you need SM3 for the same kind of reasons that others needed GOST, so it'd make sense to be consistent.

The SHA-crypt-like hashes are not only weaker against offline password cracking, but they'd also require their own optimized implementations in publicly available password crackers, or otherwise you leave this potential advantage to attackers with unpublished implementations. The yescrypt-based approach mostly avoids this because almost all processing time is spent in yescrypt, not in the wrapper function, so only the shared tasks of yescrypt optimization remains (maybe separately for defense and offense, but shared between all these wrappers).

Also, usage of these hashes (both GOST and SM3) should be discouraged in the documentation, except in cases where required.

besser82:
> Are these hashes already in use?

Huawei uses sm3crypt in EulerOS and OpenEuler; KylinOS and openKylin seem to use it, too.

solardiz:
Hashes like this (not sure if 100% compatible or not) were also used (among many others) in KoreLogic's annual Crack Me If You Can contest this summer. Support for them is also requested in https://github.com/hashcat/hashcat/issues/3987

vt-alt:
This may already be addressed, but it is worth noting that the implementation should be verified to ensure it actually matches with these distributions due to potential implementation bugs that become "standard".

besser82:
> This may already be addressed, but it is worth noting that the implementation should be verified to ensure it actually matches with these distributions due to potential implementation bugs that become "standard".

Yes, the test outputs are taken from the downstream patch in (open)Kylin:  https://gitee.com/openkylin/libxcrypt/blob/openkylin/nile/debian/patches/2-add-sm3-crypt-support.patch, which is identical to the one in EulerOS:  https://gitee.com/src-openeuler/libxcrypt/blob/master/add-sm3-crypt-support.patch

PR #200, opened 2024-12-31
NEWS: fix current version to be 4.4.38

trofi:
Without the change tarball autogeneration fails when ran as is:

    $ make dist
    ...
    NEWS not updated; not releasing


besser82:
Fixed in a2da668df3a5d8d7f05b606b6899a4ad20f990f7.

PR #201, opened 2025-01-06
Make crypt and crypt_gensalt use thread-local output buffers.

besser82:
This change makes crypt and crypt_gensalt as thread-safe as they can be without changing their interfaces.  Solaris already made this change, and it’s being discussed by glibc (with suggestion that it should be pushed upstream to the C
and POSIX standards committees): https://sourceware.org/ml/libc-alpha/2018-10/msg00437.html

Portable programs should still use the r-variants, though, because this is not a guaranteed feature that is portable to other implementations of these functions. Also it doesn’t help threads to not clobber their corresponding output buffer on a second call from within the same thread.

The way I've implemented this mimicks the implementation from Solaris.

I've also taken into account some problems with -fno-plt code and thread-local storage on PPC64-architectures.  See: https://sourceware.org/PR32387

***

This also fixes #62.

### DIFF
--- a/doc/crypt.3
+++ b/doc/crypt.3
@@ -321,6 +321,15 @@ and
 .Nm crypt_ra
 are not part of any standard.
 .Pp
+POSIX permits
+.Nm crypt
+to use a single storage area for all callers,
+even in the presence of threads.
+This library uses per-thread storage areas,
+but portable code should avoid calling
+.Nm crypt
+from multiple threads simultaneously.
+.Pp
 POSIX does not specify any hashing methods,
 and does not require hashed passphrases to be portable between systems.
 In practice, hashed passphrases are portable
@@ -392,6 +401,17 @@ uses a thread-specific static storage buffer,
 which makes it safe to call from multiple threads simultaneously,
 but does not prevent each call within a thread
 from overwriting the results of the previous one.
+If an application needs to call
+.Nm crypt
+from parallel threads independently,
+it is strongly advised to allocate thread-specific
+.Vt "struct crypt_data"
+objects from within that program itself, and to use these objects
+with one of the re-entrant interfaces provided through
+.Nm crypt_r ,
+.Nm crypt_rn ,
+or
+.Nm crypt_ra .
 .Sh BUGS
 Some implementations of
 .Nm crypt ,
@@ -433,14 +453,18 @@ l l l.
 Interface	Attribute	Value
 T{
 .Nm crypt
-T}	Thread safety	MT-Unsafe race:crypt
+T}	Thread safety	MT-Safe*
 T{
 .Nm crypt_r ,
 .Nm crypt_rn ,
 .Nm crypt_ra
 T}	Thread safety	MT-Safe
 .TE
-.sp
+.Pp
+\&* In this implementation.
+Do NOT rely on this to be portable behavior.
+See
+.Sx PORTABILITY NOTES .
 .Sh HISTORY
 A rotor-based
 .Nm crypt
--- a/doc/crypt_gensalt.3
+++ b/doc/crypt_gensalt.3
@@ -132,7 +132,9 @@ is unspecified and must not be relied upon.
 places its result in a static storage area,
 which will be overwritten by subsequent calls to
 .Nm crypt_gensalt .
-It is not safe to call
+In this version of libcrypt,
+the static storage area is allocated separately for each thread,
+but portable code should not call
 .Nm crypt_gensalt
 from multiple threads simultaneously.
 However, it
@@ -229,6 +231,22 @@ A function with the name
 .Nm crypt_gensalt
 also exists on Solaris 10 and newer, but its prototype and semantics differ.
 .Pp
+Some implementations of
+.Nm crypt_gensalt
+use a single storage area for all callers,
+even in the presence of threads.
+This library uses per-thread storage areas,
+but portable code should avoid calling
+.Nm crypt_gensalt
+from multiple threads simultaneously; instead it is strongly advised to
+allocate thread-specific buffers of
+.Dv CRYPT_GENSALT_OUTPUT_SIZE
+length from within each thread itself, and to use these buffers with the
+re-entrant interface provided through
+.Nm crypt_gensalt_rn ,
+or
+.Nm crypt_gensalt_ra .
+.Pp
 The default prefix and auto entropy features are available since libxcrypt
 version 4.0.0.  Portable software can use feature test macros to find out
 whether null pointers can be used for the
@@ -250,14 +268,18 @@ l l l.
 Interface	Attribute	Value
 T{
 .Nm crypt_gensalt
-T}	Thread safety	MT-Unsafe race:crypt_gensalt
+T}	Thread safety	MT-Safe*
 T{
 .Nm crypt_gensalt_rn ,
 .Nm crypt_gensalt_ra
 T}	Thread safety	MT-Safe
 .TE
 .ad
-.sp
+.Pp
+\&* In this implementation.
+Do NOT rely on this to be portable behavior.
+See
+.Sx PORTABILITY NOTES .
 .Sh SEE ALSO
 .Xr crypt 3 ,
 .Xr getpass 3 ,
--- a/lib/crypt-static.c
+++ b/lib/crypt-static.c
@@ -1,5 +1,5 @@
 /* Copyright (C) 2007-2017 Thorsten Kukuk
-   Copyright (C) 2019 Björn Esser
+   Copyright (C) 2019,2024 Björn Esser
 
    This library is free software; you can redistribute it and/or
    modify it under the terms of the GNU Lesser General Public License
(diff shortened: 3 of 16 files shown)


doc/crypt.3
```diff
@@ -392,6 +401,16 @@ uses a thread-specific static storage buffer,
...
+If an application needs to call
+.Nm crypt
+from parallel threads independently,
+it is strongly advised to allocate thread-specific
+.Vt "struct crypt_data"
+objects from within that program itself, and to use these objects
+with one of the re-entrant interfaces provided through
+.Nm crypt_r ,
+or
+.Nm crypt_rn .
```
solardiz, line 413: or `crypt_ra`

doc/crypt.3
```diff
@@ -433,14 +452,18 @@ l l l.
...
 T{
 .Nm crypt_r ,
 .Nm crypt_rn ,
 .Nm crypt_ra
 T}	Thread safety	MT-Safe
 .TE
-.sp
+.Pp
+\&* In this implementation.
+Do NOT rely on this to be portable behaviour.
```
solardiz, line 464: libxcrypt currently mostly uses the `behavior` spelling (not `behaviour`), so maybe it's best to use that consistently.

doc/crypt_gensalt.3
```diff
@@ -229,6 +231,20 @@ A function with the name
...
+even in the presence of threads.
+This library uses per-thread storage areas,
+but portable code should avoid calling
+.Nm crypt_gensalt
+from multiple threads simultaneously; instead it is strongly advised to
+allocate thread-specific buffers of
+.Dv CRYPT_GENSALT_OUTPUT_SIZE
+length from within each thread itself, and to use these buffers with the
+re-entrant interface provided through
+.Nm crypt_gensalt_rn .
```
solardiz, line 246: or `crypt_gensalt_ra`

doc/crypt_gensalt.3
```diff
@@ -250,14 +266,18 @@ l l l.
...
 T{
 .Nm crypt_gensalt_rn ,
 .Nm crypt_gensalt_ra
 T}	Thread safety	MT-Safe
 .TE
 .ad
-.sp
+.Pp
+\&* In this implementation.
+Do NOT rely on this to be portable behaviour.
```
solardiz, line 278: libxcrypt currently mostly uses the `behavior` spelling (not `behaviour`), so maybe it's best to use that consistently.

lib/crypt-static.c
```diff
@@ -25,8 +25,17 @@
 char *
 crypt (const char *key, const char *setting)
 {
-  static struct crypt_data nr_crypt_ctx;
-  return crypt_r (key, setting, &nr_crypt_ctx);
+  static TLS char output[CRYPT_OUTPUT_SIZE];
+  struct crypt_data nr_crypt_ctx;
+
+  crypt_r (key, setting, &nr_crypt_ctx);
```
solardiz, line 31: Now this calls `crypt_r` with uninitialized stack data in `nr_crypt_ctx` - a bug? IIRC, at least the `initialized` field should be set to 0.

Also, we probably incur some performance penalty by having this struct recreated each time, but perhaps only for `descrypt`?

solardiz (changes requested):
I didn't review the m4 stuff.

Overall, I am not sure this is worth the complexity and reduced portability, but I don't mind (once the issues I pointed out are addressed - especially what looks like a trivial yet major bug).

doc/crypt.3
besser82: Fixed.

doc/crypt_gensalt.3
besser82: Fixed.

doc/crypt_gensalt.3
besser82: Fixed.

lib/crypt-static.c
besser82: > Now this calls `crypt_r` with uninitialized stack data in `nr_crypt_ctx` - a bug? IIRC, at least the `initialized` field should be set to 0.

Fixed by initializing the whole struct with `explicit_bzero`.


> Also, we probably incur some performance penalty by having this struct recreated each time, but perhaps only for `descrypt`?

Well, the performance penalty is meassurable, but not noticeable:  This change adds on average about 220 milliseconds of computing time for 1 Million invocations; so in a range of 200 - 300 nanoseconds per call.  As those numbers are measured on an Intel Celeron 4205U with Fedora Workstation 41, the real life time penalty on affordable faster hardware, like a Ryzen 5 2400G, are expected be lower.

zackw:
@besser82 I should have time to look at this early next week, but not before that.

lib/crypt-static.c
solardiz: > Fixed by initializing the whole struct with `explicit_bzero`.

Feels like mild misuse of `explicit_bzero`, but that's not necessarily bad. If we want to use it not only to initialize, but also to wipe any sensitive data that might have been left from a previous call, then perhaps we also want to do it for the output buffer before the `strcpy_or_abort`. And then maybe move the `explicit_bzero` of the data struct to after the `strcpy_or_abort` and introduce simple initialization (just `memset`)?

besser82: `strcpy_or_abort` already zeroizes the left over space after the copied string.

The cleanup of `struct crypt_data` is also already done by `do_crypt` (called from inside `crypt_r`) after the computation has been finished.

I've used `explicit_bzero` here intentionally to ensure the just stack-allocated `struct crypt_data` is really initialized properly before use.

besser82: > Feels like mild misuse of `explicit_bzero`, but that's not necessarily bad.

Changed to `memset` for initialization.

zackw (changes requested):
I approve of the goal here.  It's something I wanted to do myself quite some time ago, in fact, but I thought there was no way to do it without making `libcrypt.so` have a hard dependency on `libpthread.so` (which is a thing we almost certainly cannot do, for compatibility reasons).  You found a way to avoid that dependency, which is great...

... except that I think the way you _did_ do it is _also_ going to be a compatibility problem.  You're using `CRYPT_OUTPUT_SIZE + CRYPT_GENSALT_OUTPUT_SIZE == 576` bytes of thread-local storage. If I'm reading [glibc's `elf/dl-tls.c`](https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/dl-tls.c;hb=HEAD) correctly, a bunch of internal data structures, and the official minimum size for thread stacks, are sized on the assumption that any given shared library is going to want only a couple of pointers' worth of TLS space.  What this means is, processes that load `libcrypt.so` are going to be pushed into a slow, poorly tested, fallback code path within the dynamic linker, and any process that loads `libcrypt.so` and also uses small thread stacks may start crashing, _whether or not they use `crypt` or `crypt_gensalt`._  On top of which, if they _do_ call `crypt`, this change makes _that_ require 32kB more stack than it used to.

I also have concerns about all the sketchy autoconf-archive code this patch imports. Just on a quick skim I saw several constructs that are asking for trouble with autoconf 2.70 and later.

----

So how _can_ we do this?

I think we shouldn't use TLS at all (which conveniently gets rid of two-thirds of the questionable autoconf macros).  I think we should use `pthread_getspecific` to acquire one thread-local pointer for each function, and allocate the bulk of the data with `malloc` (which also lets us go back to reusing `struct crypt_data` across invocations).

Unfortunately, this brings back the problem of a dependency on libpthread.  For GNU libc, libpthread was merged into libc.so.6 in glibc 2.34, and as I understand it, musl libc has always shipped a single integrated shared object for the whole C library.  But we probably still need to care about older glibc, and maybe also about non-Linux-based systems.  I think what we need to do about this is split libcrypt into four pieces that are tied back together with a `GROUP` linker script (for an example of what I mean by `GROUP` linker script, look at the contents of the file glibc installs as `libc.so` (not `libc.so.6`)):

* `libcrypt.so.3` defines _only_ `crypt_checksalt`, `crypt_gensalt_ra`, `crypt_gensalt_rn`, `crypt_preferred_method`, `crypt_ra`, and `crypt_rn`, with their `XCRYPT_x.y` symbol versions.
* `libcrypt_legacy_nonreentrant.so.1` defines _only_ `crypt` and `crypt_gensalt`, using `pthread_getspecific`-based thread-local pointers.  It has a dependency on libpthread if necessary.  We install a symlink to this file under the name `libcrypt.so.2`.
* `libcrypt_legacy_des.so.1` defines `encrypt`, `encrypt_r`, `setkey`, and `setkey_r`.  (I propose that we do not bother making `encrypt` and `setkey` use thread-local storage.)
* `libcrypt.so.1` defines all the compatibility symbols.

Thoughts?

solardiz:
> I approve of the goal here. It's something I wanted to do myself quite some time ago

If implementations look so risky and/or complicated to you, then I'm puzzled that you think the benefits outweigh the risks and effort and complexity. Maybe just don't do this?

> I think we should use `pthread_getspecific` to acquire one thread-local pointer for each function, and allocate the bulk of the data with `malloc`

This sounds somewhat similar to what I saw Solaris do, https://github.com/besser82/libxcrypt/pull/62#issuecomment-2588580961

rfc1036:
Why care about old glibc versions? It's not like people normally build libxcrypt to replace the one shipped by their distribution.

zackw:
@rfc1036 The _entire point_ of this library is to be a drop-in replacement for old glibc's libcrypt. If it didn't need to be that, it would have a completely different API.

rfc1036:
Of course, but reasonably nobody is replacing glibc's libcrypt on existing systems!

zackw:
Not true, people do it all the time, just look through the outstanding bug reports.

Issue #204, opened 2025-01-15
Is zeroing of data->internal without reset of data->initialized safe?


solardiz:
This issue is a code review in progress. I see no suitable issue label for that, so am leaving it with none.

While reviewing #201, I looked into other places where we may zeroize the data struct, and found this:

https://github.com/besser82/libxcrypt/blob/88ecfa55f8f2df0dc1fe261122afd1ba59a64aee/lib/crypt.c#L178

```c
  struct crypt_internal *cint = get_internal (data);
  h->crypt (phrase, phr_size, setting, set_size,
            (unsigned char *)data->output, sizeof data->output,
            cint->alg_specific, sizeof cint->alg_specific);

  explicit_bzero (data->internal, sizeof data->internal);
```

This zeroizes `internal`, but leaves the `initialized` field as-is. For hash implementations that compute and cache constants in `internal`, like I think `descrypt` does, doesn't this result in incorrect computation on the next call?

How does this even pass tests? Maybe I'm misinformed and our `descrypt` doesn't cache anything anymore?

Let's at least figure this out.

besser82:
AFAIK, the only purpose of the `initialized` field in `struct crypt_data` is for backwards compatibility with glibc.  I cannot find any code, that actually checks whether that variable has a particular value assigned.

Anyways, I agree upon explicitly setting this to `0` after zeroizing `internal`, in order to comply with the documentation.

solardiz:
> AFAIK, the only purpose of the `initialized` field in `struct crypt_data` is for backwards compatibility with glibc. I cannot find any code, that actually checks whether that variable has a particular value assigned.

Oh, right. I also don't see any uses of it in libxcrypt, not even in much older versions where the struct actually matched glibc's. And now that the struct doesn't even match glibc's, the only compatibility that comes from having this field is that the same source code can be built against either library - but the resulting binaries wouldn't be compatible (portable between libxcrypt and glibc libcrypt) because the struct layout is different.

PR #205, opened 2025-01-15
Fixes for Issue #204.

besser82:
* crypt: Zeroize initialized and reserved data members after computation.
* crypt: Zeroize and initialize (re)allocated memory in crypt_ra.
* test/short-outbuf: Add two more cases for crypt_ra.

### DIFF
--- a/lib/crypt.c
+++ b/lib/crypt.c
@@ -181,6 +181,8 @@ do_crypt (const char *phrase, const char *setting, struct crypt_data *data)
             cint->alg_specific, sizeof cint->alg_specific);
 
   explicit_bzero (data->internal, sizeof data->internal);
+  explicit_bzero (data->reserved, sizeof data->reserved);
+  data->initialized = 0;
 }
 
 #if INCLUDE_crypt_rn
--- a/lib/crypt.c
+++ b/lib/crypt.c
@@ -207,20 +207,23 @@ SYMVER_crypt_rn;
 char *
 crypt_ra (const char *phrase, const char *setting, void **data, int *size)
 {
-  if (!*data)
+  if (!*data || *size < 0 || (size_t) *size < sizeof (struct crypt_data))
     {
-      *data = malloc (sizeof (struct crypt_data));
-      if (!*data)
-        return 0;
-      *size = sizeof (struct crypt_data);
-    }
-  if (*size < 0 || (size_t)*size < sizeof (struct crypt_data))
-    {
+      /* realloc gives us no way to zeroize the previous data,
+         if it happens to relocate it to a new memory address.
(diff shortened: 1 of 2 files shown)


solardiz:
BTW, `realloc` gives us no way to zeroize the previous data if it happens to relocate it to a new address. This could be worth a source code comment maybe, if we started to care about these things here.

besser82:
> BTW, `realloc` gives us no way to zeroize the previous data if it happens to relocate it to a new address. This could be worth a source code comment maybe, if we started to care about these things here.

I've taken that into account in the rebased commits.

solardiz:
The code on its own looks good to me now, although I suggest to use a simple `memset` after the `realloc` because in that place it's pure initialization and not zeroization of sensitive data.

While at it, I once again notice that we have some unfortunate (in my opinion) recommendations in the man page: "Applications are encouraged, but not required, to use the phrase and setting fields to store the strings that they will pass as phrase and setting to crypt_r." I think it's very wrong that we don't just keep this struct opaque to the callers (except maybe for the `initialized` field). We may want to fix that separately.

besser82:
> The code on its own looks good to me now, although I suggest to use a simple `memset` after the `realloc` because in that place it's pure initialization and not zeroization of sensitive data.

Fixed in merged commits.


> While at it, I once again notice that we have some unfortunate (in my opinion) recommendations in the man page: "Applications are encouraged, but not required, to use the phrase and setting fields to store the strings that they will pass as phrase and setting to crypt_r." I think it's very wrong that we don't just keep this struct opaque to the callers (except maybe for the `initialized` field). We may want to fix that separately.

Let's do this in a seperate PR.

lib/crypt.c
```diff
@@ -181,6 +181,8 @@ do_crypt (const char *phrase, const char *setting, struct crypt_data *data)
             cint->alg_specific, sizeof cint->alg_specific);
 
   explicit_bzero (data->internal, sizeof data->internal);
+  explicit_bzero (data->reserved, sizeof data->reserved);
```
solardiz, line 184: I suggest we don't zeroize `reserved`. This serves no purpose with current libxcrypt, and it could unnecessarily limit how we're able to use the reserved space in future/mixed versions.

solardiz: Oh, I didn't realize you merged this already. Nevermind, then.

besser82: > I suggest we don't zeroize `reserved`. This serves no purpose with current libxcrypt, and it could unnecessarily limit how we're able to use the reserved space in future/mixed versions.

I think we can drop it, *if* we decide to use that field to store a state or sth. in later versions.

PR #206, opened 2025-02-14
Add sm3{,-yescrypt} backend and tests.

besser82:
These hashing methods are useful in applications that need modern password hashing methods, but require to rely on cryptographic properties that have been approved by the Chinese Office of State Commercial Cryptography Administration (OSCCA).

ShangMi 3 is a cryptographic hash function used in the Chinese National Standard.  It was published by the National Cryptography Administration (Chinese: 国家密码管理局) on 2010-12-17 as "GM/T 0004-2012: SM3 cryptographic hash algorithm".

SM3 is used for implementing digital signatures, message authentication codes, and pseudorandom number generators.  The algorithm is public and is considered similar to SHA-256 in security and efficiency.  SM3 is also used with Transport Layer Security.

SM3 is defined in each of:
* GM/T 0004-2012:  SM3 cryptographic hash algorithm
* GB/T 32905-2016:  Information security techniques SM3 cryptographic hash algorithm
* ISO/IEC 10118-3:2018:  IT Security techniques—Hash-functions, Part 3: Dedicated hash-functions
* IETF RFC draft-sca-cfrg-sm3-02

See:  https://datatracker.ietf.org/doc/html/draft-sca-cfrg-sm3

sm3crypt is currently in active use by the EulerOS, Kylin, openEuler, and openKylin distributions.

***

The additionally - in hope of adaption by the afore mentioned distributions - implemented sm3-yescrypt uses the output from the yescrypt hashing method in place of a hmac message.  Thus, the yescrypt crypto properties are superseeded by the ShangMi 3 hash function with a 256 bit digest.

***

Fixes #188.


doc/crypt.5
```diff
@@ -227,6 +239,16 @@ Acceptable for new hashes.
 The default processing cost parameter is 5000,
 which is too low for modern hardware.
 .hash "$5$" "\e$5\e$(rounds=[1-9][0-9]+\e$)?[^$:\(rsn]{1,16}\e$[./0-9A-Za-z]{43}" unlimited 8 256 256 "6 to 96" "1000 to 999,999,999"
+.Ss sm3crypt
+A hash based on the ShangMi 3 hash function
+with 256-bit output, that uses the same design as
+sha256crypt and/or sha512crypt.
+Supported on EulerOS, openEuler, KylinOS and openKylin,
+but not common elsewhere.
+Acceptable for new hashes.
```
solardiz: I think we should use the `Acceptable for new hashes where required.` wording also for `sm3crypt`, do not omit the ` where required` part here.

solardiz:
> The additionally - in hope of adaption by the afore mentioned distributions - implemented sm3-yescrypt

While I did suggest this in #188, that was before I learned that the existing SHA-crypt based algorithm is already widely used. I have mixed feelings about us introducing an extra algorithm, even if better, given that these are only for standards compliance in China. I understand there's value not only for Chinese users in supporting the existing hashes that they use. Also, greater (even if indirect) usage of yescrypt may result in feedback that would help us improve yescrypt itself. So I don't mind.

Regarding testing, as I recall for gost-yescrypt back then I suggested to @vt-alt to manually run tests with yescrypt hacked to always return a constant. I suggest that you try the same here. This should show two things (to manually confirm):
1. That the final hash value changes when yescrypt's return value is changed from the real hash to the constant. To confirm that the combined algorithm doesn't inadvertently ignore yescrypt.
2. That the final hash value changes with input changes (so test with multiple different inputs, separately for password and salt) even when yescrypt hash fails to change. To confirm that the outer layer does provide (at least) basic hash properties on its own.

No need to have these tests in the tree (like we don't for gost-yescrypt). This is just something to confirm manually once.

Overall, I skimmed, but I'm not eager to spend a lot of my time on this PR. Thanks!

doc/crypt.5
besser82: Thanks for the feedback!  Fixed in rebased commit.

besser82:
> Regarding testing, as I recall for gost-yescrypt back then I suggested to @vt-alt to manually run tests with yescrypt hacked to always return a constant. I suggest that you try the same here. This should show two things (to manually confirm):
> 
> 1. That the final hash value changes when yescrypt's return value is changed from the real hash to the constant. To confirm that the combined algorithm doesn't inadvertently ignore yescrypt.
> 2. That the final hash value changes with input changes (so test with multiple different inputs, separately for password and salt) even when yescrypt hash fails to change. To confirm that the outer layer does provide (at least) basic hash properties on its own.
> 
> No need to have these tests in the tree (like we don't for gost-yescrypt). This is just something to confirm manually once.

Confirming the HMAC capsule works as intended.

besser82:
For the record:  Request for hashcat supporting sm3crypt: https://github.com/hashcat/hashcat/issues/3987

thesamesam:
@besser82 We've had some requests to backport this for compat with other distros who are using it already. Could you consider a new release? Many thanks.

besser82:
> @besser82 We've had some requests to backport this for compat with other distros who are using it already. Could you consider a new release? Many thanks.

@thesamesam v4.5.0 just got released.

PR #207, opened 2025-03-09
yescrypt: Use AVX512VL XOP-like bit rotates for faster Salsa20

solardiz:
This speeds up classic scrypt by up to a third, but yescrypt only very slightly. It also reduces code size.

This would only be enabled in builds that support AVX512VL, which I guess is currently uncommon for libxcrypt (even though CPUs that support it are pretty common by now). So maybe in distros like Gentoo where packages are commonly built from source for the local machine?

It shouldn't hurt to have this in the source code anyhow, except that it'd not be part of usual coverage testing - so I suggest occasional testing of a `-march=native` build on a recent CPU, which would be a good idea prior to these changes as well.

I got the same changes into upstream yescrypt, its fork in JtR tree, and yespower earlier today. The AVX512VL code passed tests in all 3 of these on Intel Tiger Lake.

I did not actually test these changes in libxcrypt yet.

### DIFF
--- a/lib/alg-yescrypt-opt.c
+++ b/lib/alg-yescrypt-opt.c
@@ -1,6 +1,6 @@
 /*-
  * Copyright 2009 Colin Percival
- * Copyright 2012-2018 Alexander Peslyak
+ * Copyright 2012-2025 Alexander Peslyak
  * All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


besser82:
LGTM!

PR #210, opened 2025-04-11
lib/crypt.c: Do not overwrite the output buffer too early.

### LINKED ISSUE #209: crypt() overwrites its output buffer before reading arguments
As reported in https://www.openwall.com/lists/yescrypt/2025/04/08/1 passing a pointer returned by crypt(3) previously as the salt for a subsequent crypt(3) call fails. Perhaps it should succeed (arguments fully processed/copied before the output buffer is first written to) or the man page should clarify that "It's also not safe to use the pointer returned as an argument for another call to crypt(3)." or both. I suspect the same could be happening for the password argument, which could somewhat realistically happen in case a certain system uses nested password hashing (perhaps as an upgrade of legacy hashing).

besser82:
Allow passing a pointer returned by crypt(3) previously as the input for a subsequent call to crypt(3), adapt the manpage accordingly, and implement a simple testcase for such a scenario.  This also applies to the other variants of the crypt function, when the same data object is used for	subsequent calls.

Alter the resize process of the data object by crypt_ra(3), to keep the literals passed with the phrase and the setting parameter during the runtime of the function, if those are stored within the initially passed data object.

Also apply the same semantics to the crypt_gensalt(3) function family, and the internally used make_failure_token() helper.

Fixes #209.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,6 +72,7 @@ test/crypt-des
 test/crypt-gost-yescrypt
 test/crypt-kat
 test/crypt-md5
+test/crypt-nested-call
 test/crypt-nthash
 test/crypt-pbkdf1-sha1
 test/crypt-scrypt
@@ -87,6 +88,7 @@ test/explicit-bzero
 test/fcrypt-enosys
 test/gensalt
 test/gensalt-bcrypt_x
+test/gensalt-nested-call
 test/gensalt-nthash
 test/gensalt-extradata
 test/getrandom-fallbacks
--- a/Makefile.am
+++ b/Makefile.am
@@ -385,12 +385,14 @@ check_PROGRAMS = \
 	test/compile-strong-alias \
 	test/crypt-badargs \
 	test/crypt-gost-yescrypt \
+	test/crypt-nested-call \
 	test/crypt-sm3-yescrypt \
 	test/crypt-too-long-phrase \
 	test/explicit-bzero \
 	test/gensalt \
 	test/gensalt-bcrypt_x \
 	test/gensalt-extradata \
+	test/gensalt-nested-call \
 	test/gensalt-nthash \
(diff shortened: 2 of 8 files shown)


solardiz:
Thanks! I agree with the man page edits, but disagree with the code change. Let's not special-case this. There are other related cases to take care of as well: passing the previous return value as `key` rather than `setting`, and passing a pointer into the middle of a previous return value (e.g., skip over the original salt when chaining multiple hash types).

solardiz:
To be clear, I do not mean to also special-case those other cases - rather, to develop a generic fix for all of these that would not need to check any conditions. Just don't overwrite the output buffer before reading both inputs, ever.

besser82:
> To be clear, I do not mean to also special-case those other cases - rather, to develop a generic fix for all of these that would not need to check any conditions. Just don't overwrite the output buffer before reading both inputs, ever.

Should be addressed properly in the rebased commit.

besser82:
@solardiz Is everything okay with the changes now?

solardiz (approved):
The changes look OK to me now. However, I am concerned about a few related things (I'll write a separate comment).

solardiz:
I'm sorry it took me a while to get to reviewing this again. It's hard for me to make time when I can concentrate on critical code reviews now.

I'm concerned that we're adding complexity to paper over prior code's shortcomings, but OTOH I understand that we care not only about risk of bugs in latest code, but about risk of bugs across all versions, and more code changes mean increasing the latter risk. That's an unfortunate tradeoff.

For brand new code, we'd probably postpone the `make_failure_token` calls until we reach an error path.

Please don't make any changes to address this comment yet. Let's discuss first, along with the additional concern I just brought up in #209. Maybe @ldv-alt would want to weigh in?

PR #217, opened 2025-11-11
lib/crypt.h: provide list of supported hash algorithms

kotopesutility:
IMHO, it might be useful for users to get a list of hashing algorithms supported by the current build. I decided to use a dictionary-like structure so that users could match the algorithm name with the corresponding prefix.

So, I think, user can iterate over this list until they get {NULL, NULL}:
https://github.com/altlinux/pyxcrypt/commit/c56c9dba3fe6a2518db5919c1fb7e1c95c4e039b#diff-43dec635b9590d1a1e5874610e291500ffc204b411a10313a60d099106452f47R126

But they will have to check if the linked version of libxcrypt is higher than some value (where this list was implemented). Or libxcrypt can also provide some var to check with ```#ifdef```

### DIFF
--- a/test/crypt-get-supported-hash-methods.c
+++ b/test/crypt-get-supported-hash-methods.c
@@ -1,0 +1,76 @@
+#include <stdio.h>
+#include <string.h>
+
+#include "crypt-port.h"
+
+const char defined_hash_methods[] =
+#if INCLUDE_bcrypt
+  "bcrypt:$2b$;"
+#endif
+#if INCLUDE_bcrypt_a
+  "bcrypt_a:$2a$;"
+#endif
+#if INCLUDE_bcrypt_x
+  "bcrypt_x:$2x$;"
+#endif
+#if INCLUDE_bcrypt_y
+  "bcrypt_y:$2y$;"
+#endif
+#if INCLUDE_bigcrypt
+  "bigcrypt:;"
+#endif
+#if INCLUDE_bsdicrypt
+  "bsdicrypt:_;"
+#endif
+#if INCLUDE_descrypt
+  "descrypt:;"
+#endif
+#if INCLUDE_gost_yescrypt
+  "gost_yescrypt:$gy$;"
+#endif
+#if INCLUDE_md5crypt
+  "md5crypt:$1$;"
+#endif
+#if INCLUDE_nt
+  "nt:$3$;"
+#endif
+#if INCLUDE_scrypt
+  "scrypt:$7$;"
+#endif
+#if INCLUDE_sha1crypt
+  "sha1crypt:$sha1;"
+#endif
+#if INCLUDE_sha256crypt
+  "sha256crypt:$5$;"
+#endif
+#if INCLUDE_sha512crypt
+  "sha512crypt:$6$;"
+#endif
+#if INCLUDE_sm3_yescrypt
+  "sm3_yescrypt:$sm3y$;"
+#endif
+#if INCLUDE_sm3crypt
+  "sm3crypt:$sm3$;"
+#endif
+#if INCLUDE_sunmd5
+  "sunmd5:$md5;"
+#endif
+#if INCLUDE_yescrypt
+  "yescrypt:$y$;"
+#endif
+"";
+
+int main(void)
+{
+    const char *expected = defined_hash_methods;
+    const char *reported = crypt_get_supported_hash_methods();
+    size_t len1 = strlen(expected) - 1, len2 = strlen(reported);
+    int res = (len1 != len2 || strncmp(expected, reported, len2));
+
+    // Assume there is ";" in the end of defined_hash_methods
+    fprintf(stderr, "Compiled and reported by"
+            " crypt_get_supported_hash_methods() lists of supported"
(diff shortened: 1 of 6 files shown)


besser82:
Thank you for your PR!

After having a first look at this there are some things I'd like to change:

* Having a static array of supported hash algorithms in the `crypt.h` header seems a bit problematic, as that solution is not portable in such a way, that you cannot rely on this array to be accurate on run-time.  Let say you build your application against a build of libxcrypt that supports `glibc,strong,alt`, one cannot simply replace the libxcrypt library with another build that supoorts `glibc,strong,debian` for instance. The application does not know about such a change without being rebuilt.
* I'd favor a runtime evaluation with a function that returns a pointer to such an array, which is built into libxcrypt. Ideally such a function would be something like: `const hash_method * crypt_get_supported_hash_methods (size_t *num_methods)`. The presense of this function could be signaled in `crypt.h` like `#define CRYPT_GET_SUPPORTED_HASH_METHODS_AVAILABLE 1`.
* Such a function (and the underlying static array) should live inside its own compilation unit, so users of a static archive don't need to link it into their application, if it is not needed by their code.
* A good place to put the macro definition for the contents of the array would be `crypt-hashes.h`.

It would be nice if you could make these changes to your PR, and I'll review again.

vt-alt:
Perhaps, `num_methods` is not needed, because array is static. But it would not hurt either. Also, as an alternative idea, return value could be a string (of some defined, extensible format) instead of binary array. Such as, `"name1,prefix1;name2,prefix2;..."`.

kotopesutility:
> Also, as an alternative idea, return value could be a string (of some defined, extensible format) instead of binary array. Such as, `"name1,prefix1;name2,prefix2;..."`.

It makes user to parse the output of this function. IMHO, we have to do it inside this function.

besser82:
> Perhaps, `num_methods` is not needed, because array is static. But it would not hurt either. Also, as an alternative idea, return value could be a string (of some defined, extensible format) instead of binary array. Such as, `"name1,prefix1;name2,prefix2;..."`.

> > Also, as an alternative idea, return value could be a string (of some defined, extensible format) instead of binary array. Such as, `"name1,prefix1;name2,prefix2;..."`.
> 
> It makes user to parse the output of this function. IMHO, we have to do it inside this function.

I'm fine with either way, it can be a `char *` as long as we use a format that is easy to understand and fast to parse for the consumer, it can be some static binary array as well.

kotopesutility:
> > Perhaps, `num_methods` is not needed, because array is static. But it would not hurt either. Also, as an alternative idea, return value could be a string (of some defined, extensible format) instead of binary array. Such as, `"name1,prefix1;name2,prefix2;..."`.
> 
> > > Also, as an alternative idea, return value could be a string (of some defined, extensible format) instead of binary array. Such as, `"name1,prefix1;name2,prefix2;..."`.
> > 
> > 
> > It makes user to parse the output of this function. IMHO, we have to do it inside this function.
> 
> I'm fine with either way, it can be a `char *` as long as we use a format that is easy to understand and fast to parse for the consumer, it can be some static binary array as well.

I decided @vt-alt's solution is better. This evening, I'll rewrite the parsing of ```const char *crypt_get_supported_hash_methods``` output in ```pyxcrypt``` as a usage example.

kotopesutility:
@besser82, I've added simple tests for the list of provided hashing methods. I've also set the symbol version to XCRYPT_4.5, but it might be better to increase the minor version number to 6 and then set the symbol version to XCRYPT_4.6. Please, take a look

vt-alt:
Should not test/crypt-get-supported-hash-methods print something on success?

test/crypt-get-supported-hash-methods.c
```diff
@@ -0,0 +1,77 @@
...
+{
+    const char *pair_1 = defined_hash_methods;
+    const char *pair_2 = crypt_get_supported_hash_methods();
+    size_t len1 = strlen(pair_1) - 1, len2 = strlen(pair_2);
+    int res = 0;
+
+    // Assume there is ";" in the end of defined_hash_methods
+    if ((res = (len1 != len2 || strncmp(pair_1, pair_2, len2))))
+    {
+        fprintf(stderr, "Provided and real list of supported hash methods "
```
vt-alt: Perhaps, "compiled and supported ...", because "provided" and "real" looks synonymous to me. Or, s/supported/reported/ (by crypt_get_supported_hash_methods).

kotopesutility:
> Should not test/crypt-get-supported-hash-methods print something on success?


In my opinion, only if the test fails. But I think it would be possible to output something like "{provided} == {expected}." In any case, the result will be in the test/crypt-get-supported-hash-methods.log file.

test/crypt-get-supported-hash-methods.c
kotopesutility: Or maybe "expected" instead of "real"? But I like "compiled"/"supported"

kotopesutility:
@vt-alt, now it prints message both on fail/success

kotopesutility:
@besser82, is this PR suitable for you or do I need to make any adjustments?

kotopesutility:
@besser82, any updates?

Issue #214, opened 2025-10-23
test failures on powerpc-unknown-linux-gnu


NyCodeGHG:
Hello!

I'm seeing test failures while trying to natively compile on a `powerpc-unknown-linux-gnu` machine.

test-suite.log:
https://gist.github.com/NyCodeGHG/d58a69d80a134f32f376ed3b3d6ffd69

build log:
https://gist.github.com/NyCodeGHG/93531dcc1c5efb3d493ca25168abec2e

PR #219, opened 2025-11-22
test/getrandom-fallbacks: fix open shim types

nwf:
In C, functions taking a variable number of arguments are not guaranteed to have the same calling convention as functions taking a fixed number of arguments [^0].  Since `open()` and `open64()` are defined as taking a variable number of arguments, the `__wrap_open()` and `__wrap_open64()` mocks must be as well.

[^0]: C23 §6.7.7.4¶14 says: "For two function types to be compatible, both shall specify compatible return types. Moreover, the parameter type lists shall agree in the number of parameters and in use of the final ellipsis; corresponding parameters shall have compatible types."

This fixes the test on my powerpc64el machine building nixpkgs a8d610af3f1.


besser82:
Thanks for your contribution!  This should at least partially fix #214.

PR #220, opened 2025-12-10
fix -Werror=discarded-qualifiers

ep69:
On Fedora rawhide (to be Fedora 44), gcc became more strict wrt. const-ness.


besser82:
Thank you!

PR #222, opened 2026-02-11
lib/crypt-sha*, lib/crypt-sm3: fix misleading comment

RavuAlHemio:
A copied comment misrepresents the calculation step of the S byte sequence in the crypt algorithms based on SHA2 and sm3. Replace it with an actual description of what is happening.


solardiz (approved):

solardiz:
Thanks. FWIW, this is one of the most ridiculous aspects of SHA-crypt - a timing leak that exists for no good reason at all (not a tradeoff, just a ridiculous extra). I previously mentioned this in: https://github.com/besser82/libxcrypt/pull/35#issuecomment-422526061

